PR #21, opened 2015-02-09
Backend printer

bschommer:
This branch contains a single printer for the different backends which uses different modules for the different target architectures and systems.

Advantage:
* Less code duplications for syntax parts which all backends have in common
* One point for adding support for debug information or other extensions

Currently only the printer for the toplevel functions and global variables are generalized.


backend/PrintAsmaux.ml
```diff
@@ -0,0 +1,135 @@
...
+
+(** Auxiliary functions for printing of asm *)
+
+module type TARGET =
+    sig
+      val print_prologue: out_channel -> unit
+      val print_epilogue: out_channel -> unit
+      val print_align: out_channel -> int -> unit
+      val print_comm_symb:  out_channel -> Z.t -> P.t -> int -> unit
+      val print_var_info: bool
```
xavierleroy: I would prefer print_var_info and print_fun_info to be functions that the Target printer can (re)define freely.  For ELF ports these functions would just be the default functions defined here in PrintAsmaux.  For a.out ports the functions would do nothing.  My fear is that some future port may need a solution that is neither of these two.

backend/PrintAsmaux.ml
```diff
@@ -0,0 +1,135 @@
...
+
+let transl_label lbl =
+  try
+    Hashtbl.find current_function_labels lbl
+  with Not_found ->
+    let lbl' = new_label() in
+    Hashtbl.add current_function_labels lbl lbl';
+    lbl'
+
+let label oc lbl =
```
xavierleroy: This definition of "label" is appropriate for ELF targets but maybe not for all.  Since it is very short, why not leave it to each Target to define it?

backend/PrintAsmaux.ml
```diff
@@ -0,0 +1,135 @@
...
+  jumptables := []
+
+(* Variables used for the handling of varargs *)
+
+let current_function_stacksize = ref 0l
+let current_function_sig =
+  ref { sig_args = []; sig_res = None; sig_cc = cc_default }
+
+(* Functions for printing of symbol names *)
+let symbol oc symb =
```
xavierleroy: Again, these definitions are good for ELF but not for every target.  I can understand you want to avoid duplicating them in every ELF Target, but then what about calling them "elf_symbol" and "elf_symbol_offset" ?

backend/PrintAsmaux.ml
```diff
@@ -0,0 +1,135 @@
...
+let symbol_offset oc (symb, ofs) =
+  symbol oc symb;
+  let ofs = camlint_of_coqint ofs in
+  if ofs <> 0l then fprintf oc " + %ld" ofs
+
+(* The comment deliminiter *)
+let comment = "#"
+
+(* Functions for fun and var info *)
+let print_fun_info oc name =
```
xavierleroy: Again these definitions are ELF-specific.  I'd rather have them named elf_print_fun_info and elf_print_var_info.

arm/TargetPrinter.ml
```diff
@@ -0,0 +1,1140 @@
...
+    sig
+      val float_abi: float_abi_type
+      val vfpv3: bool
+      val hardware_idiv: bool
+     end
+
+(* Module containing the printing functions *)
+
+module Target (Opt: PRINTER_OPTIONS) : TARGET =
+  (struct
```
xavierleroy: The extra parenthese around "(struct ... end)" are not necessary.

backend/PrintAsmaux.ml
```diff
@@ -0,0 +1,135 @@
...
+(* Functions for printing of symbol names *)
+let symbol oc symb =
+  fprintf oc "%s" (extern_atom symb)
+
+let symbol_offset oc (symb, ofs) =
+  symbol oc symb;
+  let ofs = camlint_of_coqint ofs in
+  if ofs <> 0l then fprintf oc " + %ld" ofs
+
+(* The comment deliminiter *)
```
xavierleroy: Every Target defines its own comment character, which is quite often not "#".  It's not useful to have a default definition here.  It can also be confusing (e.g. code that would incorrectly use PrintAsmaux.comment instead of Target.comment).

ia32/TargetPrinter.ml
```diff
@@ -0,0 +1,976 @@
...
+      val print_mov_ra: out_channel -> ireg -> ident -> unit
+      val print_fun_info: bool
+      val print_var_info: bool
+      val print_epilogue: out_channel -> unit
+      val print_comm_decl: out_channel -> P.t -> Z.t -> int -> unit
+      val print_lcomm_decl: out_channel -> P.t -> Z.t -> int -> unit
+    end
+
+(* Printer functions for cygwin *)
+module Cygwin_System =
```
xavierleroy: Better to write
   module Cygwin_System : SYSTEM = struct ... end
instead of
   module Cygwin_System = (struct ... end : SYSTEM)

ia32/TargetPrinter.ml
```diff
@@ -0,0 +1,976 @@
...
+      fprintf oc "	.comm	%a, %s, %d\n" symbol name (Z.to_string sz) al
+
+    let print_lcomm_decl oc name sz al =
+      fprintf oc "	.local	%a\n" symbol name;
+      print_comm_decl oc name sz al
+
+  end:SYSTEM)
+
+(* Printer functions for ELF *)
+module ELF_System =
```
xavierleroy: Same style issue as for Cygwin_System.

ia32/TargetPrinter.ml
```diff
@@ -0,0 +1,976 @@
...
+      fprintf oc "	.comm	%a, %s, %d\n" symbol name (Z.to_string sz) al
+
+    let print_lcomm_decl oc name sz al =
+      fprintf oc "	.local	%a\n" symbol name;
+      print_comm_decl oc name sz al
+
+  end:SYSTEM)
+
+(* Printer functions for MacOS *)
+module MacOS_System =
```
xavierleroy: Same style issue as for ELF_System.

ia32/TargetPrinter.ml
```diff
@@ -0,0 +1,976 @@
...
+
+    let print_lcomm_decl oc name sz al =
+      fprintf oc "	.lcomm	%a, %s, %d\n"
+                 symbol name (Z.to_string sz) (log2 al)
+
+  end:SYSTEM)
+
+
+module Target(System: SYSTEM):TARGET =
+  (struct
```
xavierleroy: Redundant parentheses around "struct ... end"

backend/PrintAsm.ml
```diff
@@ -0,0 +1,99 @@
...
+open Asm
+open Camlcoq
+open Datatypes
+open PrintAsmaux
+open Printf
+open Sections
+open TargetPrinter
+
+module Target = (val (sel_target ()):TARGET) 
+
```
xavierleroy: This Target module is computed too early, before the command-line was parsed.  I think we need the same trick used in the target-specific printers: functorize this code over the Target, then, in print_program, apply this functor to "(val (sel_target()): TARGET)".

bschommer:
I changed the names of the functions to elf_ and functorized the code in PrintAsm.

backend/PrintAsm.ml
```diff
@@ -0,0 +1,103 @@
...
+open AST
+open Asm
+open Camlcoq
+open Datatypes
+open PrintAsmaux
+open Printf
+open Sections
+open TargetPrinter
+
+module Target = (val (sel_target ()):TARGET)
```
xavierleroy: Now this Target module is unused and you should get rid of line 23.

ia32/TargetPrinter.ml
```diff
@@ -0,0 +1,979 @@
...
+    end
+
+(* Printer functions for cygwin *)
+module Cygwin_System : SYSTEM =
+  struct
+
+    let raw_symbol oc s =
+       fprintf oc "_%s" s
+
+    let symbol = elf_symbol
```
xavierleroy: This is incorrect.  For Cygwin, symbols must be prefixed by "_", which elf_symbol doesn't do.

backend/PrintAsm.ml
```diff
@@ -0,0 +1,103 @@
...
+    let print_location oc loc =
+      if loc <> Cutil.no_loc then Target.print_file_line oc (fst loc) (snd loc)
+          
+    let print_function oc name fn =
+      Hashtbl.clear current_function_labels;
+      Target.reset_constants ();
+      let (text, lit, jmptbl) = Target.get_section_names name in
+      Target.section oc text;
+      let alignment =
+        match !Clflags.option_falignfunctions with Some n -> n | None -> 4 in
```
xavierleroy: The default alignment for the code of a function should ideally be part of the Target structure.  In the current CompCert, it is 16 for IA32 and 4 for PowerPC and ARM.  Those values are those used by GCC, and I don't know if they make a real difference performance-wise.  But in principle it is for each Target to declare its default code alignment.

PR #30, opened 2015-03-25
Removing not used hypotheses in TREE

jhjourdan:
In the module type TREE (Maps.v), there are two hypotheses that are not used anywhere. They are annoying me in Verasco. Would it be possible to remove them ?

-- 
JH Jourdan


bschommer:
I have no objections against removing the hyportheses. @xavierleroy do you have any objections?

xavierleroy:
As discussed with JHJ, I'm fine with removing these operations from the TREE interface.  Proceeding with the merge.

PR #35, opened 2015-04-01
Fix overflows in printers for clight and csyntax.

jhjourdan:
In printers for Clight and Csyntax, there are overflows for switch case labels, because of the new internal syntax of switch cases (witch are now Z instead of Int).


xavierleroy:
Well spotted, thanks.  Merging right now.

PR #39, opened 2015-04-17
Tail-recursive array initialization (for very large files)

thranur:
Small optimizations to enable CompCert to handle files with unreasonable amounts (200k+) of array initializers.

This pull request is mostly to inform of the changes I did when trying to compile large files consisting mainly in huge arrays initialized byte per byte. I do not believe this request has production-quality level (there is duplicated code, the proof could be simplified, etc.), but it might nevertheless be useful for patching (if someone else also needs to compile such unreasonably large files).

The good news is that after only these few modifications, I was able to compile source files larger than 10MB, with almost a million initializers. The compilation time is surely larger than GCC's, but the OCaml code is more time- and memory-efficient than I expected for extracted Coq code.

### DIFF
--- a/cfrontend/C2C.ml
+++ b/cfrontend/C2C.ml
@@ -1098,27 +1098,32 @@ let convertFundecl env (sto, id, ty, optinit) =
 
 (** Initializers *)
 
+let listToInitializer l =
+  List.fold_left (fun acc e -> Init_cons(e, acc)) Init_nil l
+
 let rec convertInit env init =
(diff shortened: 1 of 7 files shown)


xavierleroy:
Thanks for the interesting suggestion.  We'll keep this idea in mind (and this pull request open) for when we run into issues with large initializers.  CompCert's non-free licensing conditions make it impossible to accept anonymous contributions to the code base.

PR #49, opened 2015-07-07
Change the definition of Typles.tuple

jhjourdan:
This new definition generates fewer universe constraints, so that it can be used in conjunction with Robbert Krebbers's CH2O semantics of C11.


xavierleroy:
This is black magic!  But indeed the new definition makes no difference for CompCert itself (in Coq 8.4 as well as 8.5, if I understand correctly), and is admittedly more legible than the previous one, so let's use it.

PR #54, opened 2015-09-21
Atomic builtins

bschommer:
This pull request makes the following atomic builtins of the gcc available for the 32bit integer ppc backend:
* __sync_fetch_and_add
* __atomic_compare_exchange
* __atomic_exchange
* __atomic_load

The implementation follows mainly the one outlined at [C/C++11 mappings to processors](https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html).
Currently only the __ATOMIC_SEQ_CST memory order is supported.

In order to avoid name clashes with the gcc builtins which additionally take a memory order argument, or in the case of atomic_compare_exchange two arguments they are prefixed by __builtin.

### DIFF
--- a/backend/RTL.v
+++ b/backend/RTL.v
@@ -104,8 +104,7 @@ Record function: Type := mkfunction {
     for its stack-allocated activation record.  [fn_params] is the list
     of registers that are bound to the values of arguments at call time.
     [fn_entrypoint] is the node of the first instruction of the function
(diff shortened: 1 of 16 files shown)


jeehoonkang:
This is really an insignificant comment, but the link to the "C/C++11 mappings to processors" is slightly broken :-)

bschommer:
The link should work now.

xavierleroy:
OK for merging.

PR #55, opened 2015-09-22
Upgrade to Flocq 2.5.0.

silene:
Note: this version of Flocq is compatible with both Coq 8.4 and 8.5.


xavierleroy:
Thanks, Guillaume.  AbsInt and I need to decide whether this code goes in the imminent next AbsInt release or just after.  Will merge pretty soon in either case.

PR #56, opened 2015-09-28
Debug locations

bschommer:
This pull request adds a new interface for the generation of dwarf debugging information. 

The information needed for the generation of the debugging information is now collected during the translation and stored in DebugInformations.ml. Later the functions in Dwarfgen.ml generate the debugging entries and locations which are then printed in the DwarfPrinter.ml.

Also the main debug functions are now all available through Debug.ml and set according the command line flag.


cfrontend/C2C.ml
```diff
@@ -524,6 +524,11 @@ let convertField env f =
   (intern_string f.fld_name, convertTyp env f.fld_typ)
 
 let convertCompositedef env su id attr members =
+  let t = match su with C.Struct -> 
+    let layout = Cutil.struct_layout env members in
+    List.iter (fun (a,b) -> Debug.set_member_offset id a b) layout;
+    TStruct (id,attr) | C.Union -> TUnion (id,attr) in
```
xavierleroy: Very hard to read because of the lack of indentation for the two cases (Struct/Union).

powerpc/Asmexpand.ml
```diff
@@ -586,22 +586,158 @@ let expand_instruction instr =
...
+    | BA_addrglobal _
+    | BA_loadstack _ -> None
+    | BA_addrstack ofs -> Some (BA_addrstack ofs)
+    | BA_splitlong (hi,lo) -> 
+        begin
+          match (aux hi,aux lo) with
+          | Some hi ,Some lo -> Some (BA_splitlong (hi,lo))
+          | _,_ -> None
+        end in
+  aux (List.hd a)
```
xavierleroy: Normally "a" should not be the emply list here, but just to make the code more robust, you could return None in the case a = [].  Or iterate "aux" over the elements of a and stop when "aux" returns Some.

powerpc/Asmexpand.ml
```diff
@@ -586,22 +586,158 @@ let expand_instruction instr =
...
+
+let expand_scope id lbl oldscopes newscopes =
+  let opening = List.filter (fun a -> not (List.mem a oldscopes)) newscopes
+  and closing = List.filter (fun a -> not (List.mem a newscopes)) oldscopes in
+  List.iter (fun i -> Debug.open_scope id i lbl) opening;
+  List.iter (fun i -> Debug.close_scope id i lbl) closing
+
+let expand_instruction id l =
+  let get_lbl = function
```
xavierleroy: At first sight, the logic of this function looks independent of the target processor.  So maybe, in the future, if we support debugging info for other platforms, this logic can be moved to a target-independent file.  But for the time being it's fine.

PR #57, opened 2015-09-30
Correction of a few bugs in the pre-parser, added comments.

jhjourdan:
Fixed a few bugs in the pre parser. In particular, the following code was not parsed correctly:

typedef int a;

int f() {
  for(int a; ;)
    if(1);
  a * x;
}

Additionally, I tried to add some comments in the pre-parser code,
especially for the different hacks used to solve various conflicts.


bschommer:
Thanks a lot for this fix.

PR #58, opened 2015-10-07
Pass --no-stdlib and -v to menhir when compiling pre_parser.mly.

fpottier:
Passing --no-stdlib ensures that there is no dependency on Menhir's
standard library.
Passing -v, which is equivalent to --explain --dump, requests the
generation of pre_parser.automaton, a description of the automaton.

General comment:
In agreement with Xavier Leroy and Jacques-Henri Jourdan, I have been
working on an experimental approach to improving the syntax error messages
in CompCert's pre_parser. For the moment, I am planning to make a series of
pull requests that tend to clean up or improve pre_parser.mly, without affecting
its behavior. This tiny pull request is mostly a way of checking that I understand
how pull requests work. Thanks!

Francois Pottier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ cparser/Parser.v
 cparser/Lexer.ml
 cparser/pre_parser.ml
 cparser/pre_parser.mli
+cparser/pre_parser.automaton
 lib/Readconfig.ml
 lib/Tokenize.ml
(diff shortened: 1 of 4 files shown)


bschommer:
Merged into trunk.

fpottier:
Thanks!

PR #60, opened 2015-10-23
New system for reporting syntax errors

fpottier:
Hello,
As the title suggests, this is a whole new system for reporting syntax errors in the pre_parser.
It requires the development version of Menhir, which is here:
  http://gallium.inria.fr/~fpottier/menhir/menhir-20151023.tar.gz
and should also reach opam soon.
I have tried to create a reasonably clean line of commits.
In summary:
  - a series of commits make changes in pre_parser.mly, without changing the language that is accepted;
  - a few commits make small changes or cleanups in Parse.ml and Lexer.mll;
  - then, a big commit introduces several new files (in particular, the database of error messages, and the GNUmakefile that helps maintain it) and plugs them in Lexer.mll.

I had some merge problems at the end that I did not quite understand; sorry for the noise.

I hope you will find this interesting and worth integrating.
Please feel free to ask questions.
Francois.


bschommer:
I just tested it against our Testsuite and everything looks fine. I have just a short questions:
Is it possible to generate the error Database before the actual compile step?

Concerning menhir: Is there any official way to hand in a patch,  since I added a simple Makefile to build menhir under Windows without ocamlbuild.

fpottier:
Dear Bernhard,

On Mon, Oct 26, 2015 at 03:19:51AM -0700, Bernhard Schommer wrote:
> Is it possible to generate the error Database before the actual compile step?

Are you referring to the creation of cparser/pre_parser_messages.ml?  At the
moment it is done as part of the "depend" target in Makefile.extr.  I am not
sure how one could do it earlier. What do you have in mind?  Of course one
could register cparser/pre_parser_messages.ml as a source file, distribute it
with CompCert, and manually re-generate it when needed. So people who want to
compile CompCert wouldn't need to generate it. But then there would be a risk
of us forgetting to re-generate it after changing pre_parser.mly.

> Concerning menhir: Is there any official way to hand in a patch, since I
> added a simple Makefile to build menhir under Windows without ocamlbuild.

Currently there isn't (Menhir's source code is hosted in the INRIA forge and
is not publicly visible). The simplest thing would be for you to send me the
Makefile, which I could then ship with Menhir.

What's the motivation for not requiring ocamlbuild?
Isn't ocamlbuild installed at the same time as ocaml?

By the way, I don't know if this is relevant, but it should soon be possible
to install menhir via opam under windows. (At least cygwin/windows.) A user
has recently suggested a patch that should allow this.

Best regards,

-- 
François Pottier
Francois.Pottier@inria.fr
http://gallium.inria.fr/~fpottier/

bschommer:
Dear François,

>Are you referring to the creation of cparser/pre_parser_messages.ml?  At the
moment it is done as part of the "depend" target in Makefile.extr.  I am not
sure how one could do it earlier. What do you have in mind?  Of course one
could register cparser/pre_parser_messages.ml as a source file, distribute it
with CompCert, and manually re-generate it when needed. So people who want to
compile CompCert wouldn't need to generate it. But then there would be a risk
of us forgetting to re-generate it after changing pre_parser.mly.

Having it as a part of the depend is sufficient for me. We currently generate the dependency file as well as the extraction files in a preprocessing step under linux and only perform the compile steps under windows.

>What's the motivation for not requiring ocamlbuild?
Isn't ocamlbuild installed at the same time as ocaml?

Unfortunately ocamlbuild does not really work on the native windows port since it requires some kind of bash under windows (cygwin or mingw) and since neither using cygwin nor mingw is an option for us.

Best regards,
Bernhard

fpottier:
On Mon, Oct 26, 2015 at 06:31:21AM -0700, Bernhard Schommer wrote:
> Having it as a part of the depend is sufficient for me.

Fine, let's not change anything, then.

> Unfortunately ocamlbuild does not really work on the native windows port
> since it requires some kind of bash under windows (cygwin or mingw) and
> since neither using cygwin nor mingw is an option for us.

I see. I would be happy to integrate your Makefile in Menhir's distribution.

-- 
François Pottier
Francois.Pottier@inria.fr
http://gallium.inria.fr/~fpottier/

PR #69, opened 2015-11-07
Parser : duplicate identifier tokens, fix K&R definition parsing

jhjourdan:
This PR has two stages:

- Another fix to the problem of the interaction between the lexer hack and the lookahead. The new solution involves emitting two tokens for each identifier. The first one is eaten by the lookahead, and only the second one contains the var_name vs. type_name information. This makes the information stored in the second token up-to-date with respect to the environment.

- Remodeling of the parsing of function definitions, so that:
  * K&R function definition s are correctly parsed, even if they return a pointer to a function
  * the grammar is closer to the standard


jhjourdan:
I forgot to say: this has an other advantage compared to the previous approach: the contexts are stored in the parsing stack, and not in a separate stack in the lexer, so that there is no longer any risk that the contexts stack is desynchronized with the parsing stack.

PR #74, opened 2015-11-23
Fix a typo in a syntax error message.

fpottier:


bschommer:
Thanks again for the fix.

PR #75, opened 2015-11-24
Simple mcpu

bschommer:
This branch provides a new command line option mcpu for specifying the e5500 as target processor instead of using the model configuration e5500.

### DIFF
--- a/configure
+++ b/configure
@@ -39,7 +39,6 @@ Supported targets:
 
 For PowerPC targets, the "ppc-" prefix can be refined into:
   ppc64-           PowerPC 64 bits
-  e5500-           Freescale e5500 core (PowerPC 64 bits + EREF extensions)
 
 For ARM targets, the "arm-" prefix can be refined into:
   armv6-           ARMv6   + VFPv2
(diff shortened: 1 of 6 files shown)


configure
```diff
@@ -88,11 +87,10 @@ struct_passing=""
 struct_return=""
 
 case "$target" in
-  powerpc-*|ppc-*|powerpc64-*|ppc64-*|e5500-*)
```
xavierleroy: Why not keeping the possibility to configure for e5500?  I mean, this would set "e5500" as the default value for the -mcpu flag.

driver/Driver.ml
```diff
@@ -574,6 +574,11 @@ let cmdline_actions =
 (* Target processor options *)
   Exact "-mthumb", Set option_mthumb;
   Exact "-marm", Unset option_mthumb;
+  Prefix "-mcpu=", Self (fun s -> let s = String.sub s 6 ((String.length s) -6) in
```
xavierleroy: Here, by symmetry, I would support both "-mcpu=ppc64" and "-mcpu=e5500".

driver/Clflags.ml
```diff
@@ -33,6 +33,7 @@ let option_falignfunctions = ref (None: int option)
 let option_falignbranchtargets = ref 0
 let option_faligncondbranchs = ref 0
 let option_finline_asm = ref false
+let option_mcpu = ref "generic"
```
xavierleroy: Following on my previous proposal, option_mcpu would be initialized to Configuration.model.

PR #79, opened 2015-12-11
Config option

bschommer:
This PR allows it to use relative paths in the .ini files for tools and libraries. The paths are relative to the given .ini file.

It also introduces a new option `-target` that allows it to select .ini files with the given name.

### DIFF
--- a/driver/Configuration.ml
+++ b/driver/Configuration.ml
@@ -10,6 +10,7 @@
 (*                                                                     *)
 (* *********************************************************************)
 
+open Filename
 open Printf
 
 let search_argv key =
@@ -22,8 +23,8 @@ let search_argv key =
   !res
 
 let absolute_path base file =
-  if Filename.is_relative file then
-    Filename.concat base file
+  if is_relative file then
+    concat base file
   else
     file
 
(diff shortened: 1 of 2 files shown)


driver/Configuration.ml
```diff
@@ -10,6 +10,7 @@
 (*                                                                     *)
 (* *********************************************************************)
 
+open Filename
```
xavierleroy: Cosmetic: I prefer not to open Filename and qualify explicitly.  E.g. "Filename.concat" is more readable than "concat" (which could be some other kind of concatenation operation).

driver/Configuration.ml
```diff
@@ -82,9 +94,19 @@ let get_config_list key =
...
   | vl -> vl
 
-let prepro = get_config_list "prepro"
-let asm = get_config_list "asm"
-let linker = get_config_list "linker"
+let tool_absolute_path tools =
+  match tools with
+  | [] -> []
+  | tool::args -> let tool =
+      if is_implicit tool && dirname tool = current_dir_name then
```
xavierleroy: I had the impression that "Filename.is_implicit tool" suffices here.  In other words, isn't it the case that "dirname tool = current_dir_name" as soon as "is_implicit tool = true" ?

bschommer: `Filename.is_implict "a/b"` returns true as well, but `Filename.dirname "a/b"` is equal to `a`.

xavierleroy:
Looks fine to me.  Sorry for  the delay.  Merging!

PR #84, opened 2016-01-12
Update cminor support for single-precision floats

jrrk2:
The cminor parser and associated support files have become out of date
with respect to the support for single-precision floats in CompCert-2.6 onwards.
Also dumping .rtl was producing previously invalid ops for not/mulhs/mulhu.

### DIFF
--- a/backend/CMlexer.mll
+++ b/backend/CMlexer.mll
@@ -25,6 +25,7 @@ let floatlit =
  ("-"? (['0'-'9'] ['0'-'9' '_']*
   ('.' ['0'-'9' '_']* )?
   (['e' 'E'] ['+' '-']? ['0'-'9'] ['0'-'9' '_']*)? )) | "inf" | "nan"
+let floatlithex = "0X" ['0'-'9' 'a'-'f' 'A'-'F']+
 let ident = ['A'-'Z' 'a'-'z' '_'] ['A'-'Z' 'a'-'z' '_' '$' '0'-'9']*
 let qident = '\'' [ ^ '\'' ]+ '\''
 let temp = "$" ['1'-'9'] ['0'-'9']*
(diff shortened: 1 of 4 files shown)


xavierleroy:
Thanks for this pull request.  The update to the x86 ops printer is welcome.  The support for single-FP ops in the Cminor parser is good too, even though in the long term this parser will probably go away and be replaced by a JSON input.

The one point I find questionable is the printing of FP literals in raw hexadecimal format for -dcminor.  For one thing, this is inconsistent with the Cminor parser.  For another thing, this is inconsistent with the printers for other intermediate languages of CompCert, making it harder to trace the compilation of a given file through all the passes.

jrrk2:
Dear Xavier,
   When I made the pull request, I was not aware of the severe loss of 
accuracy in printing floating point values in Cminor programs.
I have since made a matching change in the Cminor parser which is not 
checked in yet.
To be honest, I did not realize that making further changes later 
modifies the pull request.

As a matter of policy, in a verified compiler I think you want:

(PrintCminor (ParseCminor prog) == prog)
(ParseCminor (PrintCminor prog) == prog)

whereas you actually have:

(PrintCminor (ParseCminor prog) == prog)
(ParseCminor (PrintCminor prog) != prog)

My change to hex numbers achieves this at the expense of loss 
human-readability. You could get a compromise
capability by merely extending the precision of the printed numbers, for 
example change %F to %.15F.
This would maintain backward compatibility of the parser, which is 
actually not difficult to change anyway.

As an alternative, an undocumented flag could enable the accurate mode 
of operation.

I have plans to use cminor as a front-end to various analysis and 
conversion tools. For this I need to maintain
floating-point accuracy, there was never an intention to diverge the 
project or dilute your achievement in any way.

Regards,
Dr Jonathan Kimmitt
Research Associate
Computational Chemistry
University of Cambridge

xavierleroy:
I merged most of this PR in commits d58c35e (printing of IA32 operators) and 2f31ded (Cminor parsing of single FP operations).  I left out the raw hex format for FP literals, for reasons explained above, but increased the number of significant digits printed for FP literals in debug outputs, commit 4b0f05c.

PR #86, opened 2016-02-05
Better treatment of names in the clightgen tool

xavierleroy:
Following discussions with the VST team, here are some improvements to the clightgen tool (the tool that produces Clight AST in Coq syntax from C source files).

The main improvement is that compiler-introduced temporary variables are given names `_t'1`, `_t'2`, etc.  Before,  they were represented by integers `NNN%positive`.  The names are more convenient and more stable w.r.t. small changes in the C source file.

The SimplExpr pass was changed accordingly so as to restart the name generator at `next_unused_ident` for every function, instead of threading a name generator through all the functions of a C source file.


bschommer:
Ok for us.

PR #90, opened 2016-03-02
Make casts of pointers to _Bool semantically well defined

xavierleroy:
In CompCert 2.5 the semantics of pointer comparisons against the NULL pointer was made more accurate by making it undefined if the pointer is invalid (outside bounds).  (See pull request https://github.com/AbsInt/CompCert/pull/31).  Technical difficulties prevented this change from being propagated to the semantics of casts from pointer types to the `_Bool` type, which involves an implicit pointer comparison against NULL.  Hence, this kind of casts was temporarily given undefined semantics.

This commit makes pointer-to-`_Bool` casts semantically defined (again), provided the pointer is valid.  This reinstates the equivalence between casts to `_Bool` and comparisons `!= 0`.  The technical difficulties mentioned above came from the translation of assignments in a value context (e.g. `1 + (x = y * 2)`) in the SimplExpr pass.  The pass was lightly modified to work around the issue.

The impact is small on the operational semantics of CompCert C and Clight: the `sem_cast` function from Cop now takes the current memory state as an extra argument (needed to check the validity of a pointer value when casting to `_Bool`).  By transitivity, several other semantic definitions (such as `Cop.sem_add`) also get an extra argument of type memory state because they call `sem_cast` internally.

I'm about to ask on the compcert-users list for review and feedback on this PR, so please do not merge it immediately.

### DIFF
--- a/cfrontend/Cexec.v
+++ b/cfrontend/Cexec.v
@@ -645,11 +645,11 @@ Section EXPRS.
 Variable e: env.
 Variable w: world.
 
-Fixpoint sem_cast_arguments (vtl: list (val * type)) (tl: typelist) : option (list val) :=
+Fixpoint sem_cast_arguments (vtl: list (val * type)) (tl: typelist) (m: mem) : option (list val) :=
   match vtl, tl with
   | nil, Tnil => Some nil
   | (v1,t1)::vtl, Tcons t1' tl =>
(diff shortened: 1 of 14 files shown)


xavierleroy:
The feedback I received so far is positive.  In particular, it seems that no major change is expected in order to adapt the VST program logic to this change.  Merging right now.

PR #91, opened 2016-03-03
Add the mixed simulation reasoning principle

jeehoonkang:
I heard from @gilhur that @xavierleroy and @gilhur discussed about mixing the forward and backward reasoning principles for the simulation relation.  Here is the implementation which we call the "mixed simulation".

I do not expect this PR to be merged.  I just would like to share the implementation to those who are interested in CompCert development and compiler verification in general.  I hope there will be fruitful discussions around this PR!
## 

The mixed simulation reasoning principle lets you (locally) do forward
reasoning even if the target language as a whole is not determinate.
It is required only for the target _state_, not the entire language, to
be determinate.  This principle logically subsumes both the forawrd
and backward reasoning principle.

This principle is theoretically motivated by:

  Neis et al.  Pilsner: A Compositionally Verified Compiler for a Higher-Order Imperative Language.  ICFP 2015.

The main motivation for this principle is to fully support integer-
pointer casts in CompCert.  The following paper introduces such a model,
yet the model is nondeterministic and is not fit for the existing proof
structure of CompCert based on the forward simulation reasoning
principle.   The mixed simulation reasoning principle solves the problem
by supporting both the backward reasoning and the forward reasoning (for
determinate states).

  Kang et al.  A Formal C Memory Model Supporting Integer-Pointer Casts.  PLDI 2015.

The second motivastion is to remove axioms on external calls.  CompCert
assumes that external calls are (1) "receptive" and (2) "determinate"
(variants of determinism; defined in common/Smallstep.v).  These axioms
are introduced to enable us to use forward reasoning.  We want to lift
this restriction by using mixed simulation, or in other words, backward
reasoning for external calls and forward reasoning for the other steps.

### DIFF
--- a/powerpc/Asm.v
+++ b/powerpc/Asm.v
@@ -606,7 +606,7 @@ Definition load1 (chunk: memory_chunk) (rd: preg)
 
 Definition load2 (chunk: memory_chunk) (rd: preg) (r1 r2: ireg)
                  (rs: regset) (m: mem) :=
-  match Mem.loadv chunk m (Val.add rs#r1 rs#r2) with
+  match Mem.loadv chunk m (Val.add (gpr_or_zero rs r1) rs#r2) with
   | None => Stuck
   | Some v => Next (nextinstr (rs#rd <- v)) m
   end.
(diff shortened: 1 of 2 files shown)


xavierleroy:
Thanks, Jeehoon, for this implementation of the "mixed simulation" principle.  I haven't had time to study it in details yet, but will.

PR #92, opened 2016-03-11
Cleanup

bschommer:
This PR activates more warnings and turns all warnings into errors. Also some unused function, variables and types are removes.


xavierleroy:
For the record, after initial discussion with @bschommer:

Many of the cleanups are well taken, thanks for looking into this.  However, I'm uneasy with the changes that aim at silencing warning number 27 ("innocuous unused variable"), because I feel they diminish the readability of the code.  Compare:
```
  let rec mktable n = function
     | [] -> assert false
     | [_, _] -> []
     | (key, _) :: rem -> ...
```
and
```
 let rec mktable n = function
     | [] -> assert false
     | [key, action] -> []
     | (key, action) :: rem -> ...
```
The latter makes it clear that this function works over a list of (key, action) pairs; the former does not.  In other words, naming parts of patterns with variables that are not used is a form of code documentation, and I'd be sad to lose it.

A few more comments to follow on specific changes.

backend/IRC.ml
```diff
@@ -459,44 +449,7 @@ let initialNodePartition g =
           DLinkNode.insert n g.simplifyWorklist
     | Colored -> ()
     | _ -> assert false in
-  Hashtbl.iter part_node g.varTable
-
-
-(* Check invariants *)
```
xavierleroy: These functions (up to and including `checkInvariants`) are useful to debug the register allocator.  I'd like to keep them but it's OK to comment them out.

cfrontend/C2C.ml
```diff
@@ -76,13 +76,13 @@ let atom_sections a =
   with Not_found ->
     []
 
-let atom_is_small_data a ofs =
```
xavierleroy: The reason for passing the offset to "atom_is_small_data" is that, conceivably, "global + small offset" could be addressable via the SDA while "global + very large offset" might not, requiring normal addressing.  So, the axiomatization of SDA addressing in powerpc/Asm.v takes this possibility into account.  In reality, the compiler has no good way to know whether the offset causes SDA addressing to fail or not, which is why the "ofs" argument is actually ignored.

bschommer:
I reverted the changes introduced by warning 27 and added the invariantes checks in IRC again. I discovered a few more things we may want to cleanup and/or document (especially in the debug information).

xavierleroy:
Thanks a lot for the revert.  The cleanup looks good to me now.  There will probably be a conflict with pull request #93 (because of the refactoring of types Clight.program and Csyntax.program) but nothing serious.

bschommer:
Cleanup is now at the moment finished.

bschommer:
Resolved the merge conflicts.

xavierleroy:
Merging into master right now.

xavierleroy:
I'll let you delete the branch when you feel like it.

PR #96, opened 2016-03-24
Added the _Generic expression.

bschommer:
Adds the C11 _Generic expression, which allows chosing an expression according to the type of another expression.


bschommer:
Maybe in the future.

PR #100, opened 2016-05-24
Driver refactoring

bschommer:
Pulls out common functionality from the driver and as well as the frontend functions from the driver into its own file.


xavierleroy:
OK with me.

I'm not a big fan of the "backslash-n-backslash-newline-backslash-space" idiom, but I understand that naked newlines in string literals can behave strangely under Windows, so it's OK.

PR #102, opened 2016-06-22
Stricter control of permissions in memory injections and extensions

xavierleroy:
This pull request strengthens memory injections and extensions so as to guarantee that the permissions of existing memory locations are not increased by the injection/extension.  The only increase of permissions permitted is empty locations in the source memory state of the injection/extension being mapped to nonempty locations.  (Theorems <code>perm_extends_inv</code> and <code>perm_inject_inv</code> in module <code>Memory</code>.)

This stricter policy on permissions goes through CompCert's proofs easily, and would help with the developments of the VST project.  I'm asking the VST people for a review; don't merge immediately.


xavierleroy:
Positive feedback from the VST project.  Merging.

PR #104, opened 2016-06-30
Port trunk to Coq 8.5

jhjourdan:
This pull request ports Compcert to Coq 8.5pl1. Appart from minor tactics incompatibilities, the main changes are:

- The use of Parameter/Axiom instead of Variable/Hypothesis in module signatures
- Change some patterns: in 8.5, constructors used in patterns respect the impliciteness of parameters (this was not the case before).


xavierleroy:
Thanks a million!  I'm ready to merge (manually, so as to produce a single commit) but I'll wait for a few days for release 8.5pl2 of Coq.

jhjourdan:
Then, maybe we should better check that there is no incompatibility with the upcoming 8.5pl2 release (i.e., with the current 8.5 branch).

xavierleroy:
Manually merged into master, commit e73d5db.

PR #105, opened 2016-07-01
Fix parsing and handling of CMinor files

m-schmidt:
Currently CompCert cannot parse its .cm output since lines like the following lead to parse errors:

> extern "__i64_dtos" = runtime "__i64_dtos" : float -> long

After adding support for "runtime" to the parser build still fail because the driver was missing the ASMExpand step while handling .cm input files.

### DIFF
--- a/backend/CMlexer.mll
+++ b/backend/CMlexer.mll
@@ -136,6 +136,7 @@ rule token = parse
   | "]"    { RBRACKET }
   | "readonly" { READONLY }
   | "return"    { RETURN }
+  | "runtime"    { RUNTIME }
   | ")"    { RPAREN }
   | ";"    { SEMICOLON }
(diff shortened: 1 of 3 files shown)


m-schmidt:
I'm not sure, whether one wants a 'runtime' token in the CM parser/lexer and make it part of the 'eftoks' rule or not. If not 'runtime' in the declarations will be a matched identifier. Both ways work.

xavierleroy:
Merged, thanks.  

What this shows is that the Cminor input facility of CompCert is no longer used, and certainly no longer tested.  We should discuss whether we want to keep it alive (perhaps with a JSON or S-exp concrete syntax so that the parser can be autogenerated from the Cminor type definitions) or remove it altogether.

PR #106, opened 2016-07-05
Extend Asmexpand to handle more cases of BA_splitlong

m-schmidt:
This PR is a partial fix (currently only ia32) for the Asmexpand problem in code like this:

 volatile long long x;
 void f(unsigned int i) { x = i; }


xavierleroy:
@m-schmidt : I think it's better to avoid generating these patterns than to handle them via special-cases in Asmexpand.ml.  I'll propose a fix along this line when the migration to Coq 8.5 is over.

m-schmidt:
OK. For "expand_builtin_vstore_common" on PPC and ARM one would need one more temporary register for the expanded code.

xavierleroy:
Fixed in commit 4c7650c.

PR #107, opened 2016-07-18
Response File support

bschommer:
This pull request adds response file support to avoid problems with command line limits under Windows. A response file allows it to pass command line arguments per file. Options in the file are separated by whitespace and inserted in place.

The quoting conventions for options follows that of the gnu-tools: Options containing whitespace must be surrounded by either double or single quotes.

Additionally responsefiles are also used to pass options to the assembler, linker and preprocessor.


xavierleroy:
I agree we need to support response files for the reasons you give.  Thanks for initiating this effort.

There are a number of delicate points to address:
* Diab 5 tools use the command-line syntax `-@responsefile` while Microsoft tools and GNU tools under Win32 use `@responsefile`.  This must be taken into account when ccomp passes arguments via response files.  Probably this should be configurable via compcert.ini.
* For the various tools called from CompCert, the syntax of response files (quoting conventions, etc) needs to be fully understood.  
  * For Microsoft tools, https://msdn.microsoft.com/en-us/library/3te4xt0y.aspx suggests that response files are parsed exactly like command-line arguments, and the parsing conventions for the latter are here: https://msdn.microsoft.com/en-us/library/a1y7w461.aspx .  Note that response files cannot be nested.
  * For GNU tools, it seems like singe and double quotes are recognized, and backslash systematically escapes the next character (in any context).  No mention of nested response files.
  * For Diab 5 tools, no info is given in the manual.
* As a consequence of the above, `Filename.quote` may not produce syntax that the tool can parse.  E.g. on Win32, the output of `Filename.quote` is OK for the Microsoft tools but not for the GNU tools.
* We also need to agree on a syntax for the response files given to CompCert.  The syntax implemented in this PR (based on an earlier proposal of mine) is Unix-shell-like with the minimal modifications to ensure that the output of `Filename.quote` can be read back both on Unix and Windows.  However, under Win32 it is not compatible with the Microsoft parsing conventions (e.g. `\\` outside a quote is parsed as a single backslash, while it would denote two backslashes for a Microsoft tool).  It is not compatible with the GNU tools either (e.g. no backslash escapes between single quotes).

In a nutshell: it's complicated!  But let's try to find some answers nonetheless.

driver/Assembler.ml
```diff
@@ -18,12 +18,17 @@ open Driveraux
...
+  let cmd,opts = match Configuration.asm with
+  | name::opts -> name,opts
+  | [] -> assert false (* Should be catched in Configuration *) in
+  let opts = List.concat [
+    opts;
     ["-o"; ofile];
     List.rev !assembler_options;
     [ifile]
   ] in
+  let opts = responsefile opts (fun a -> if gnu_system then ["@"^a] else ["@"^a]) in
```
xavierleroy: Did you mean `"-@"^a` in the `else` branch?  At any rate, it might be better to take the appropriate prefix `@` or `-@` from an entry in `compcert.ini`

driver/Driveraux.ml
```diff
@@ -14,12 +14,33 @@
...
+
 (* Invocation of external tools *)
 
 let rec waitpid_no_intr pid =
   try Unix.waitpid [] pid
   with Unix.Unix_error (Unix.EINTR, _, _) -> waitpid_no_intr pid
 
+let responsefile args resp_arg =
+  try
+    if Sys.win32 && (String.length (String.concat "" args) > 7000) then
```
xavierleroy: Could do `List.fold_left (fun len arg -> len + String.length arg + 1) 0 args > 7000` here, to avoid constructing the concatenated string.  Also, one space should be counted between every argument, I think.  Finally, I vaguely remember that Cygwin has its own limitations on the command-line length, not as strict as Win32's, but not as big as Linux or MacOS.

driver/Driveraux.ml
```diff
@@ -14,12 +14,33 @@
...
+      let file,oc = Filename.open_temp_file "compcert" "" in
+      let whitespace = Str.regexp "[ \t\r\n]" in
+      let quote arg =
+        if Str.string_match whitespace arg 0 then
```
xavierleroy: Depending on the syntax for responsefiles implemented by the tool, other characters may need quoting.  E.g. backslashes with the GNU tool syntax.

driver/Frontend.ml
```diff
@@ -24,15 +24,20 @@ open Printf
...
+  let opts = List.concat [
+    opts;
     ["-D__COMPCERT__"];
     (if !Clflags.use_standard_headers
      then ["-I" ^ Filename.concat !Clflags.stdlib_path "include" ]
      else []);
     List.rev !prepro_options;
     [ifile]
   ] in
+  let opts = responsefile opts (fun a -> if gnu_system then ["@"^a] else ["@"^a]) in
```
xavierleroy: Same comment as above.

driver/Linker.ml
```diff
@@ -19,14 +19,19 @@ open Driveraux
...
+    opts;
     ["-o"; exe_name];
     files;
     (if Configuration.has_runtime_lib
-     then ["-L" ^ !stdlib_path; "-lcompcert"]
-     else [])
+    then ["-L" ^ !stdlib_path; "-lcompcert"]
+    else [])
   ] in
+  let opts = responsefile opts (fun a -> if gnu_system then ["@"^a] else ["-@"^a]) in
```
xavierleroy: Same comment as above.

driver/Assembler.ml
xavierleroy: Also: I wonder if this logic (of using a response file or not) could be factored out by putting it in `Driveraux.command` or maybe a new function `Driveraux.command_respfile` on top of `Driveraux.command`.

driver/Driveraux.ml
```diff
@@ -14,12 +14,33 @@
...
+          Filename.quote arg (* We need to quote arguments containing whitespaces *)
+        else
+          arg in
+      List.iter (fun a -> Printf.fprintf oc "%s " (quote a)) args;
+      close_out oc;
+      resp_arg file
+    else
+      args
+  with Sys_error _ ->
+    args
```
xavierleroy: When is the temporary response file removed?

bschommer:
The responsefile functionality is now merged into command and also the temporary file is removed after the command returns.

Concerning quoting: My suggestion would be to adopt the conventions of the gnu toolchain since most build systems will have support for this. I will further investigate about the exact conventions.

xavierleroy:
I agree that for generating @files and passing them to the linker, etc, we must support the GNU syntax, because most of the time it's a GNU toolchain we're using.  But I'm afraid we have to support the Diab syntax as well, which mean reverse-engineering it first...

For parsing @files given to CompCert, we could use the GNU syntax also.  My only reservation is that this syntax is not friendly towards Windows file names (backslash needs escaping, always).

driver/Commandline.ml
```diff
@@ -99,4 +100,9 @@ let parse_array spec argv first last =
   in parse first
 
 let parse_cmdline spec =
-  parse_array spec Sys.argv 1 (Array.length Sys.argv - 1)
+  try
+    let argv = expandargv Sys.argv in
+    parse_array spec argv 1 (Array.length argv - 1)
+  with Arg.Bad s ->
```
xavierleroy: Can the `Arg.Bad` exception be raised here?  I don't think so.  Perhaps you meant `Responsefile.Error`.

driver/Driveraux.ml
```diff
@@ -14,13 +14,20 @@
 open Printf
 open Clflags
 
+(* Is this a gnu based tool chain *)
+let gnu_system = Configuration.system <> "diab"
+
+(* Sage removale of files *)
```
xavierleroy: Two typos in this comment.

driver/Driveraux.ml
```diff
@@ -51,12 +58,55 @@ let command ?stdout args =
             argv.(0) fn (Unix.error_message err) param;
     -1
 
+(* This function reimplements quoting of writeargv from libiberty *)
+let gnu_quote arg =
```
xavierleroy: I'd be tempted to move `gnu_quote` and `diab_quote` to the new lib/Responsefile module.  It makes sense to have parsing and unparsing in the same module.

driver/Driveraux.ml
```diff
@@ -51,12 +58,55 @@ let command ?stdout args =
...
+    | Configuration.Unsupported -> assert false
+    | Configuration.Gnu -> gnu_quote
+    | Configuration.Diab -> diab_quote in
+    let file,oc = Filename.open_temp_file "compcert" "" in
+    let cmd,args = match args with
+    | cmd::args -> cmd,args
+    | [] -> assert false (* Should never happen *) in
+    List.iter (fun a -> Printf.fprintf oc "%s " (quote a)) args;
+    close_out oc;
+    let arg = if gnu_system then "@"^file else "-@"^file in
```
xavierleroy: It is not consistent to hardwire the choice between `@` and `-@` in the Caml code, while leaving the choice of response file syntax to the `compcert.ini` file.  I mean, either both parameters (response file syntax and prefix to put before the response file name) should be entries of `compcert.ini`, or none.

driver/Linker.ml
```diff
@@ -24,8 +24,8 @@ let linker exe_name files =
     ["-o"; exe_name];
     files;
     (if Configuration.has_runtime_lib
-     then ["-L" ^ !stdlib_path; "-lcompcert"]
-     else [])
+    then ["-L" ^ !stdlib_path; "-lcompcert"]
```
xavierleroy: It's good to avoid whitespace-only changes.  Plus, the previous indentation is better than the new one.

lib/Responsefile.mll
```diff
@@ -0,0 +1,97 @@
...
+(*  Automatique.  All rights reserved.  This file is distributed       *)
+(*  under the terms of the GNU General Public License as published by  *)
+(*  the Free Software Foundation, either version 2 of the License, or  *)
+(*  (at your option) any later version.  This file is also distributed *)
+(*  under the terms of the INRIA Non-Commercial License Agreement.     *)
+(*                                                                     *)
+(* *********************************************************************)
+
+
+(* Parsing response files with various quoting styles *)
```
xavierleroy: This comment no longer matches the contents of the file...  Better remove it.

bschommer:
I fixed the things mentioned in the comments and moved the quoting functionality into Responsefile.ml.
The responsfile style now also influences whether `@` or `-@` is used.

xavierleroy:
Thanks for the updates.  Just merged.  

Tiny comment: you probably no longer need `let gnu_system = ...` in Driveraux.ml.

PR #112, opened 2016-08-16
driver/Interp: handle the return value of `extcall printf` correctly

bpowers:
As noted here:
https://twitter.com/zhendongsu/status/765470502546571265/photo/1

### DIFF
--- a/driver/Interp.ml
+++ b/driver/Interp.ml
@@ -387,10 +387,12 @@ let do_external_function id sg ge w args m =
   match camlstring_of_coqstring id, args with
   | "printf", Vptr(b, ofs) :: args' ->
       extract_string m b ofs >>= fun fmt ->
-      print_string (do_printf m fmt args');
+      let formatted = do_printf m fmt args' in
+      let len = coqint_of_camlint (Int32.of_int (String.length formatted)) in
+      print_string formatted;
       flush stdout;
       convert_external_args ge args sg.sig_args >>= fun eargs ->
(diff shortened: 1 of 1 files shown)


bpowers:
The contortions to get an Int.int are sort of gross, and I'm not very familiar with the codebase.  Let me know if there is a better way to structure this!

driver/Interp.ml
```diff
@@ -387,10 +387,12 @@ let do_external_function id sg ge w args m =
   match camlstring_of_coqstring id, args with
   | "printf", Vptr(b, ofs) :: args' ->
       extract_string m b ofs >>= fun fmt ->
-      print_string (do_printf m fmt args');
+      let formatted = do_printf m fmt args' in
+      let len = (coqint_of_camlint (Int32.of_int (String.length formatted))) in
```
xavierleroy: The outer parentheses `(coqint_of_camlint ...)` are not needed.  Otherwise, yes, that's a decent (if not super nice) way of getting the length as a 32-bit Coq integer.

bpowers:
@xavierleroy good catch, I've updated the patch and removed them

bschommer:
I just pushed a variant of your proposed fix with the commit b08fd2ea809b87af8551ff6bc50e544209798d24. Thanks for investigating.

Also this is not a bug in the semantic since CompCert does not provide a c library and the interpreter only emulates printf in order to allow some kind of output for testing.

bpowers:
thanks for the fix and clarification @bschommer !!

PR #117, opened 2016-08-17
Support for ARM Big Endian

m-schmidt:
These are the changes from my EndiannessPlayground-branch:

- Parameter 'big_endian'-flag in Arm/Archi.v depending on ini-file
- Implemented new calling conventions (register pairs must be switched) and updated target printer
- Updated sources of runtime library for new calling conventions
- Updated configure script to allow selection of arm endianness via target triples


m-schmidt:
Wrong direction...

PR #142, opened 2016-09-21
Fix minor issues in some proofs and tactics.

maximedenes:
These minor problems were revealed by porting CompCert to Coq 8.6, where
they trigger errors.

All these fixes are backward compatible and have been tested with Coq 8.5pl2.


xavierleroy:
Thanks, Maxime, for the fixes.  The Ltac scripts in Cexec are not pretty indeed and would deserve a rewrite, but this will keep them working for some more time.

PR #145, opened 2016-10-02
Support for 64-bit target processors

xavierleroy:
This pull request introduces CompCert support for 64-bit target processors and (as a first instantiation) extends the x86/ia32 port of CompCert to generate 64-bit code.  (Configure for `x86_64-linux`.)

This is a large, invasive change in CompCert.  Some design guidelines and general explanations are given below.  To facilitate review, this PR is split in 3 commits:
1. Generic support for 64-bit and 32-bit target processors in the target-independent parts of CompCert.  This is the biggest set of changes (about +4500 -2500 lines).
2. Updating the PowerPC port.  This port remains 32-bit only, but some work is needed to adapt it to the changes above.  The amount of work is moderate: about +500 -300 lines.  Similar work will need to be done on the ARM port.
3. Updating the IA32/x86 port and adding 64-bit support to it.  This is the first 64-bit port of CompCert, and actually it is a combined 32- and 64-bit port, sharing lots of code between the two modes.  Unsurprisingly, it is large: +5000 -1600 lines.

## Motivations

Currently, CompCert generates code for 32-bit architectures.  In particular, the hypothesis that a pointer fits in 4 bytes is hard-wired in many places.  

There is some demand for supporting 64-bit architectures in CompCert, for the usual reasons:
* Addressing more than 4 Gb of memory.
* Faster arithmetic over 64-bit integers.
* Compatibility with OS and precompiled libraries that are 64-bit only.

It would be relatively easy to convert CompCert to 64-bit target architectures only, dropping support for 32-bit targets.  However, most critical embedded systems still use 32-bit processors.  The
challenge, therefore, is to support both 32-bit and 64-bit target architectures in the same CompCert code base.

## 64-bit integer arithmetic

CompCert 2.0 introduced support for 64-bit integers in addition to 32-bit integers, with correct handling of mixed-size arithmetic (combining a 32-bit argument and a 64-bit argument).  However, the back-end implementation of 64-bit integers and their arithmetic operation is specific to 32-bit architectures: 64-bit integers are split in two 32-bit halves, each half being held in a 32-bit integer register or in a stack location.  Arithmetic and logical operations over those split integers are implemented during the Selection phase, with help from some built-in functions and library functions.

CompCert 2.4 and CompCert 2.6 make it possible to have integer registers that are 64-bit wide and not just 32-bit wide.  The last step for full support of a 64-bit processor is to implement 64-bit integer arithmetic operations by generating the corresponding processor instructions, instead of emulating them via pairs of 32-bit integers.

In this PR, this goal is achieved by having a parameter `Archi.splitlong` that indicates whether 64-bit integers must be handled as pairs of 32-bit integers, or whether the processor supports 64-bit registers and 64-bit arithmetic operations.  In the latter case, the processor description in `$ARCHI` provides a `SelectLong` module that does custom instruction selection for 64-bit integer operators.  In the former case, this `$ARCHI/SelectLong` module just reexports the generic instruction selection based on splitting in two 32-bit halves, which is now in `backend/Splitlong`.  Dual-bitsize ports such as the new IA32/x86 port select between the two approaches at run-time depending on the selected target.

## 64-bit pointers and pointer arithmetic

The treatment of pointers is more problematic.  Pointer values are currently defined as
```
         Vptr: block -> int -> value
```
with the offset part being a 32-bit integer (type `int`).  This makes sense for a 32-bit architecture: the maximal size of a memory block is 2<sup>32</sup> bytes, hence a 32-bit unsigned integer is adequate to address a byte within a block.  For a 64-bit architecture, we want to support memory blocks larger than 2<sup>32</sup> bytes, and to be able to address every byte of those big blocks.  Hence, the offset part of pointer values must be a 64-bit integer:
<pre>
         Vptr: block -> int64 -> value
</pre>

How to support these two models in the same code base?  One early idea that led nowhere so far is to use unbounded mathematical integers '''Z''' instead of machine integers `int` or `int64` as offsets in `Vptr` values.  The approach followed in this PR introduces a third integer type `ptrofs` to denote offsets in pointer values:
<pre>
         Vptr: block -> ptrofs -> value
</pre>
Just like `int` and `int64` and their operations are obtained by instantiating a generic functor `Integers.Make` over the desired bit sizes (32 and 64, respectively), `ptrofs` is obtained by instantiating `Integers.Make` with size
```
        if Archi.ptr64 then 64%nat else 32%nat
```
(Here and elsewhere, `Archi.ptr64` is an architectural parameter indicating the size of pointers.)  This gives us a "fresh" `ptrofs` type that is always distinct from `int` and from `int64`, but is isomorphic to one of those two types depending on the value of `Archi.ptr64`.  We capture this isomorphism by defining conversion functions between `ptrofs`, `int` and `int64` and by showing properties of these functions such as
<pre>
        forall n: int64, Archi.ptr64 = true -> Ptrofs.to_int64 (Ptrofs.of_int64 n) = n.
</pre>

Using this `ptrofs` abstraction in the CompCert C semantics is relatively natural.  For example, pointer plus integer addition is now defined as: (see `Cop.sem_add`)
* convert the integer argument (of size 32 or 64 bits depending on its type) to a `ptrofs` value;
* add this `ptrofs` to the offset part of the `Vptr` value of the pointer argument.

This subsumes the current, 32-bit-only definition of pointer plus integer addition, namely:
* pointer + 32-bit integer: add integer to offset part of pointer
* pointer + 64-bit integer: cast integer to 32 bits and add to offset part of pointer.

We get the same behavior by using `ptrofs` in the case `Archi.ptr64 = false`.  In the other case, we get the behavior we expect on a 64-bit platform:
* pointer + 32-bit integer: cast integer to 64 bits and add to offset part of pointer
* pointer + 64-bit integer: add integer offset part of pointer.

## Pointers as integers, integers as pointers

In many places of the current CompCert, `Vptr` pointer values and `Vint` 32-bit integer values can be used interchangeably.  This is a conscious design decision, and it helps with:
* handling the `intptr_t` and `uintptr_t` types of the C standard;
* handling the null pointer like the integer zero;
* using the same processor instructions (say, `add` on PowerPC or `lea` on x86) for integer arithmetic and for pointer arithmetic, thus reflecting what actually goes on in the processor once pointers and integers are both bit patterns.

For example, the `Val.add` operation on values, which models the 32-bit integer add instruction of the processor, is
```
Definition add (v1 v2: val): val :=
  match v1, v2 with
  | Vint n1, Vint n2 => Vint(Int.add n1 n2)
  | Vptr b1 ofs1, Vint n2 => Vptr b1 (Int.add ofs1 n2)
  | Vint n1, Vptr b2 ofs2 => Vptr b2 (Int.add ofs2 n1)
  | _, _ => Vundef
  end.
```
Addition of a pointer and a 64-bit integer is not defined at the value level; instead, the CompCert C semantics treats it as the addition of the pointer and the low 32 bits of the integer.  

For a 64-bit target architecture, it's now `Vlong` 64-bit integer values that can be used in the same places as `Vptr` pointer values (as much as possible).  Hence, 32-bit integer addition `Val.add` should no longer accept `Vptr` arguments, but 64-bit integer addition `Val.addl` must accept them!

The approach followed in this PR is to make these operations over values dependent on the `Archi.ptr64` parameter, e.g.:
```
Definition add (v1 v2: val): val :=
  match v1, v2 with
  | Vint n1, Vint n2 => Vint(Int.add n1 n2)
  | Vptr b1 ofs1, Vint n2 => if Archi.ptr64 then Vundef else Vptr b1 (Ptrofs.add ofs1 (Ptrofs.of_int n2))
  | Vint n1, Vptr b2 ofs2 => if Archi.ptr64 then Vundef else Vptr b2 (Ptrofs.add ofs2 (Ptrofs.of_int n1))
  | _, _ => Vundef
  end.

Definition addl (v1 v2: val): val :=
  match v1, v2 with
  | Vlong n1, Vlong n2 => Vlong(Int64.add n1 n2)
  | Vptr b1 ofs1, Vlong n2 => if Archi.ptr64 then Vptr b1 (Ptrofs.add ofs1 (Ptrofs.of_int64 n2)) else Vundef
  | Vlong n1, Vptr b2 ofs2 => if Archi.ptr64 then Vptr b2 (Ptrofs.add ofs2 (Ptrofs.of_int64 n1)) else Vundef
  | _, _ => Vundef
  end.
```

It works, but reasoning over these definitions is quite heavy.

## size_t and ptrdiff_t

It's not just the type `intptr_t` that is 32 or 64-bit wide depending on the target platform, but also the types
* `size_t` for the result of `sizeof` and for the size argument of `malloc`
* `ptrdiff_t` for the result of pointer - pointer subtraction.

Values of those types are either `Vint n` (for a 32-bit target) or `Vlong n` (for a 64-bit target).  This difference can be abstracted over to some extent by defining
```
        Definition Vptrofs (n: ptrofs) :=
          if Archi.ptr64 then Vlong (Ptrofs.to_int64 n) else Vint (Ptrofs.to_int n).
```
and computing sizes and pointer differences at type `ptrofs` before injecting them into values via `Vptrofs`.

## Limitations and future work

Concerning the general approach:
* Fewer optimizations are performed on 64-bit integer arithmetic than on 32-bit integer arithmetic.  This is the case for neededness analysis and its use for dead code elimination.  This could be improved.
* Likewise, 64-bit integer division could be optimized more.
* Currently, loading a small integer (say, 8 unsigned bits) as a 64-bit integer involves two operations: a load as a 32-bit integer followed by zero-extension to 64 bits.  To do better we need to introduce new memory chunks such as `Oint8unsigned64` i.e. "a 8-bit unsigned quantity loaded as / stored from a 64-bit integer".

Concerning the x86-64 port in particular:
* Several points of non-conformance with the ELF ABI: calling conventions for passing structs by value; 16-alignment of some composite types when they are large enough.
* The regression tests in `test/regression` need updating.


robbertkrebbers:
This looks great! You write:

>  One early idea that led nowhere so far is to use unbounded mathematical integers `Z` instead of machine integer

Any intuition why it led nowhere?

xavierleroy:
@robbertkrebbers : pointer offsets in Z work quite nicely in the CompCert C / Clight semantics, as you could have guessed.  It's further down the compiler that things get weird.  For example, at C level, the natural semantics for pointer + integer addition is:
```
     p : T * = Vptr b ofs
     i : unsigned int = Vint delta
    ------------------------------------
     p + i : T * = Vptr b (ofs + Int.unsigned delta * sizeof T)%Z
```
but we cannot compile this down to Cminor code like `p + i * sizeof(T)` because the multiplication is an integer multiplication modulo 2<sup>32</sup> and it would compute a different pointer than the one predicted above.  So, we'd need pseudo-machine instructions such as "multiply unsigned integer by constant then add to pointer" to capture the right semantics, and would not be able to optimize them as much as normal integer arithmetic (replacing multiplications by shifts, reassociate additions, etc).  And these pseudo-machine instructions would be with us all the way down to Asm.

Another way to state my problem with this "pointer offsets in Z" approach is that it invalidates some algebraic properties of pointer arithmetic that are not guaranteed by the C standard but nonetheless important for compilation, such as:
```
      p + (i1 + i2) == (p + i1) + i2                if p is a pointer and i1, i2 are integers
      p + i == (T *) ((char *) p + i * sizeof(T))   if p has type pointer to T
```

xavierleroy:
Feedback is positive, the earlier 32-bit ports seem to be working without regressions, and the new x86-64 port is basically usable although it needs more work (64-bit-specific optimizations, ABI conformance, etc).  Therefore, I am merging this branch in master.

PR #146, opened 2016-10-15
Anonymous Structs

bschommer:
This branch implements support for the C11 feature of anonymous composite members.

### DIFF
--- a/cfrontend/C2C.ml
+++ b/cfrontend/C2C.ml
@@ -533,13 +533,13 @@ let rec convertTypArgs env tl el =
 let convertField env f =
   if f.fld_bitfield <> None then
     unsupported "bit field in struct or union (consider adding option -fbitfields)";
-  (intern_string f.fld_name, convertTyp env f.fld_typ)
+  (intern_string f.fld_name.name, convertTyp env f.fld_typ)
 
 let convertCompositedef env su id attr members =
   let t = match su with
   | C.Struct ->
       let layout = Cutil.struct_layout env members in
-      List.iter (fun (a,b) -> Debug.set_member_offset id a b) layout;
+      List.iter (fun (a,b) -> Debug.set_member_offset id a.name b) layout;
       TStruct (id,attr)
   | C.Union -> TUnion (id,attr) in
   Debug.set_composite_size id su (Cutil.sizeof env t);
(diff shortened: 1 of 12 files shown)


xavierleroy:
Interesting experiment, thanks a lot.  Here are a few issues I spotted.  Consider:
```
struct s {
  struct { int x, y; };
  struct { double u, v; };
};
```
The C2C translation fails to cope with more than one anonymous member, because all accesses with field name the empty string go to the first anonymous member.  For example, the code below results in a retyping error during C2C:
```
double proj_u (struct s * p) { return (*p).u; }
```

Probably easier to fix: the `->` notation also causes retyping errors:
```
int proj_y (struct s * p) { return (*p).y; }  // works
int proj_x (struct s * p) { return p->x; }    // doesn't work
```
Finally the pretty printing of elaborated C code needs fixing to avoid multiple dots as in
```
int proj_y(struct s * p)
{
  return (*p)..y;
}
```

bschommer:
The problem with the retyping and the printing should be solved by renaming the anonymous members.

bschommer:
Arrow needs investigation.

bschommer:
Arrow case now also works as intended.

xavierleroy:
I did some more tests.  Ambiguous declarations, where the same name can refer to several different members, are sometimes detected but not always:
```
struct s {
  struct { int x, y; };
  char * x;
};                           // detected: error: duplicate member 'x'

struct s {
  char * x;
  struct { int x, y; };
};                           // not detected

struct s {
  struct { int x, y; };
  struct { double x, v; };
};                           // not detected
```

In the "not detected" cases, the name `x` is resolved sometimes to the last declaration (in textual order) and sometimes to the first.

Also: the warning "anonymous structs/unions are a C11 extension" should probably be turned off, at least by default.

bschommer:
I extended the duplication check. Concerning the "anonymous structs/unions are a C11 extension": I would like to keep it on since the default language supported by CompCert is C99 and if someone wants to use C11 features one either has to explicitly disable the warnings for C11 or life with them.

xavierleroy:
More tests ahead.  CompCert accepts the code below, while GCC and Clang (even in C11 mode) reject it:
```
struct S1 { int x; };

struct S2 { struct S1; };

int get_x(struct S2 * s2) { return s2->x; }
```
GCC's diagnostics:
```
/tmp/bar.c:3:22: warning: declaration does not declare anything
/tmp/bar.c: In function ‘get_x’:
/tmp/bar.c:5:38: error: ‘struct S2’ has no member named ‘x’
```

xavierleroy:
The following is a name collision issue, so it depends on which version and target of CompCert is used.  But you get the idea:
```
struct S {
  int _126;   // choose name that will cause collision with the name generated for the anonymous member
  struct { int y; };
};

int get_y(struct S * s) { return s->y; }
```
This fails on a retyping error:
```
/tmp/foo.c:6: error: retyping error: argument of .y is not a struct or union
/tmp/foo.c:6: error: retyping error: not a l-value
```
The more I think about it, the more I believe idents should not be used to name members of structs and unions: 1- idents are designed to support alpha-conversion, while members are not alpha-convertible, and 2- idents give a sense of uniqueness that can be misleading, as in the case above.  I'd rather keep members as strings and implement a specific algorithm to pick fresh names for anonymous members.

bschommer:
Beside the other problems:
   
    struct S1 { int x; };
    
    struct S2 { struct S1; };
    
should not trigger a anonymous struct warning but as gcc/clang point out that the declaration does not declare anything, even if anonymous structs are not supported.

bschommer:
I opened a new PR which creates names instead of using idents.

PR #150, opened 2016-11-22
porting to 8.6. Warning: an Admitted in Deadcodeproof.

MarisaKirisame:
Hello, I am trying to port compcert to 8.6. 
I need some help here, as I have a admit in backend/Deadcodeproof.v that I cant get rid off... 
Also, I havent test whether it work or not on 8.5 (but I'm pretty sure it dont, because some tactic are now 'stronger' and automatically do more, like backend/Unusedglobproof.v, and some name changes), so what could be done to keep the compatibility?

### DIFF
--- a/backend/Deadcodeproof.v
+++ b/backend/Deadcodeproof.v
@@ -978,11 +978,13 @@ Ltac UseTransfer :=
   eapply match_succ_states; eauto. simpl; auto.
   destruct res; auto. apply eagree_set_undef; auto.
   eapply magree_storebytes_left; eauto.
-  exploit aaddr_arg_sound; eauto. 
-  intros (bc & A & B & C).
-  intros. eapply nlive_contains; eauto.
+  exploit aaddr_arg_sound; eauto.
+  intros (bc & A & B & C); intros.
   erewrite Mem.loadbytes_length in H0 by eauto.
-  rewrite nat_of_Z_eq in H0 by omega. auto.
+  rewrite nat_of_Z_eq in H0 by omega.
+  eapply nlive_contains; eauto.
+  subst adst.
+  admit.
 + (* annot *)
   destruct (transfer_builtin_args (kill_builtin_res res ne, nm) _x1) as (ne1, nm1) eqn:TR.
   InvSoundState.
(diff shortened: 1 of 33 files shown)


MarisaKirisame:
also intro x; do_something x break in cparser/Parser.v, but it seems to be generated... can anyone hand me a pointer to how vy work?

xavierleroy:
Concerning the problem with Coq 8.6 and the Menhir-generated file Parser.v, the expert is @jhjourdan and I hope he'll be able to look into that at some future point.

Concerning compatibility in general, every major version of Coq breaks CompCert, so we don't even try to support multiple versions of Coq: every release of CompCert is pinned to one version of Coq.  

So, some future release of CompCert will be for Coq 8.6, but I don't know which one yet and at what time we will make the switch to 8.6, since it is often necessary to wait for a pl1 or pl2 bug-fix release of Coq.

jhjourdan:
Indeed, I'll have a look at that. What is the priority?

MarisaKirisame:
The menhir problem lies not in compcert but in menhir source code...
https://github.com/npm-ml/menhir/blob/19e70f27c4bb3b99a01d8cdff6e301bde74d4fba/src/coqBackend.ml
at line 247 and 412 the line 
fprintf f "Local Obligation Tactic := intro x; case x; reflexivity.\n\n";
break in 8.6. 
simply changing it to
fprintf f "Local Obligation Tactic := let x := fresh in intro x; case x; reflexivity.\n\n";
seems to be able to fix the problem. I will make a pull request to menhir soon...

jhjourdan:
Nice, thank you for the patch. It should be compatible with 8.6.

I will patch menhir soon and @fpottier should be able to do a release.

jhjourdan:
I did that on menhir. It worked. If you need that, I guess  @fpottier can do a release soon.

Btw, there are still quite a few errors/warnings shown by Coq.

MarisaKirisame:
@jhjourdan most of the warnings are Implicit Arguments being deprecated; I will fix that soon.
However I cannot discharge the admit in deadcode... Also what other errors do you got?

MarisaKirisame:
The latest commit remove some of the warning

xavierleroy:
Thanks for your contribution, it's great to have others exploring new Coq versions :-)  See also pull request #156 for a complementary effort.  I'm discussing with the other parties involved to decide when to switch CompCert to Coq 8.6.

xavierleroy:
The switch to Coq 8.6 is complete, both in git master and in the latest release 3.0.1.  Thanks for all your input.

MarisaKirisame:
👍

PR #153, opened 2016-12-07
Next try for support of anonymous structs.

bschommer:
Instead of using idents the anonymous fileds get names of the
for <anon>_c where c is a counter of all anonymous members.

### DIFF
--- a/cparser/Elab.ml
+++ b/cparser/Elab.ml
@@ -760,11 +760,10 @@ and elab_init_name_group keep_ty loc env (spec, namelist) =
 
 and elab_field_group keep_ty env (Field_group (spec, fieldlist, loc)) =
 
-  let fieldlist = List.map (function
-    | (None, x) ->
-      (Name ("", JUSTBASE, [], loc), x)
-    | (Some n, x) -> (n, x))
-      fieldlist
+  let fieldlist = List.map
+    (function (None, x) -> (Name ("", JUSTBASE, [], loc), x)
+            | (Some n, x) -> (n, x))
+    fieldlist
   in
 
   let ((names, env'), sto) =
--- a/cparser/Elab.ml
+++ b/cparser/Elab.ml
@@ -1179,7 +1179,19 @@ module I = struct
         end else
           None
     | _, _ ->
-        None
+      None
+
+  let has_member env name ty =
+    let mem f id =
+      try
+        ignore(f env (id,name)); true
+      with _ -> false in
+    match ty with
+    | TStruct (id,_) ->
+      mem Env.find_struct_member id
+    | TUnion (id,_) ->
+      mem Env.find_union_member id
+    | _ -> false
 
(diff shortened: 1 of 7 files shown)


xavierleroy:
I like the new code very much.  However, there is still a problem with multiple anonymous members in the same struct (my first example in #146 ).  Remember: counters are meant to be incremented :-)

If we wanted the options `-dc` and `-dclight` to produce textual CompCert C / Clight that is also valid C source code, we would have to invent better field names than `<anon>__`NNN.  However, I don't think we want right now, and this part of the implementation (picking fresh names) can always be changed later.

A couple more remarks in the code review to follow.

cparser/Elab.ml
```diff
@@ -760,15 +760,17 @@ and elab_init_name_group keep_ty loc env (spec, namelist) =
...
-      | (None, x) -> (Name ("", JUSTBASE, [], loc), x)
-      | (Some n, x) -> (n, x))
-    fieldlist
+  let fieldlist,_ = mmap (fun c id ->
+    match id with
+    | (None, x) ->
+      let name = Printf.sprintf "<anon>__%d" c in
+      (Name (name, JUSTBASE, [], loc), x,true),c+1
+    | (Some n, x) -> (n, x,false),c)
+      0 fieldlist
```
xavierleroy: This generates unique names within a field group but not within a complete struct or union declaration.

My suggestion would be to use the empty string as name, then in `elab_struct_or_union_info`, onlce the list `m` of members is obtained, scan it again to replace empty names with fresh names.

cparser/Elab.ml
```diff
@@ -1427,9 +1442,14 @@ let elab_expr vararg loc env a =
         | _ ->
             error "request for member '%s' in something not a structure or union" fieldname in
       (* A field of a const/volatile struct or union is itself const/volatile *)
-      { edesc = EUnop(Odot fieldname, b1);
-        etyp = add_attributes_type (List.filter attr_inherited_by_members attrs)
-                                   (type_of_member env fld) },env
+      let rec aux = function
```
xavierleroy: Nondescriptive function names such as `aux` are not nice.  What about `access`? or `accessor`?  or `make_access`?  or...

cparser/Elab.ml
```diff
@@ -1446,9 +1466,18 @@ let elab_expr vararg loc env a =
             end
         | _ ->
             error "member reference type %a is not a pointer" (print_typ env) b1.etyp in
-      { edesc = EUnop(Oarrow fieldname, b1);
-        etyp = add_attributes_type (List.filter attr_inherited_by_members attrs)
-                                   (type_of_member env fld) },env
+       let rec aux =  function
```
xavierleroy: Same comment as above for the name `aux`

xavierleroy:
I reviewed again and pushed two very minor edits (see commits).

I think this is good to go!  Feel free to merge at your leisure, or give me green light to merge.

PR #156, opened 2017-01-09
Coq 8.6 port

maximedenes:
How about starting the new year by taking the plunge and moving to Coq 8.6?

Only a few small changes were required (they are not backward compatible, though).

I know that the usual policy of CompCert would be to wait for 8.6.1, but so far users have reported good stability, few incompatibilities and nice performance improvements, so you might want to give it a try. In any case, feel free to do as you like, this PR will at least give a way to experiment with CompCert under Coq 8.6.

I started working on this patch before #150 was submitted, so there is some duplication, sorry about that. I'm not fixing any warnings here, so it may still be worth to later cherry-pick some changes from #150 on top of this one.

### DIFF
--- a/backend/Deadcodeproof.v
+++ b/backend/Deadcodeproof.v
@@ -978,7 +978,8 @@ Ltac UseTransfer :=
   eapply match_succ_states; eauto. simpl; auto.
   destruct res; auto. apply eagree_set_undef; auto.
   eapply magree_storebytes_left; eauto.
-  exploit aaddr_arg_sound; eauto. 
+  clear H3.
+  exploit aaddr_arg_sound; eauto.
   intros (bc & A & B & C).
   intros. eapply nlive_contains; eauto.
   erewrite Mem.loadbytes_length in H0 by eauto.
--- a/backend/Inliningproof.v
+++ b/backend/Inliningproof.v
@@ -1171,11 +1171,11 @@ Proof.
   rewrite <- SP in MS0.
   eapply match_stacks_invariant; eauto.
     intros. destruct (eq_block b1 stk).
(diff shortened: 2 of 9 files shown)


xavierleroy:
Thanks a lot @maximedenes for this (other) 8.6 port.  I'm discussing with other parties to decide when to switch CompCert to 8.6.  

I note that Flocq's latest release is compatible with 8.5 and 8.6, so as a first move I'll update CompCert's copy of Flocq very soon.

jhjourdan:
By the way, now that opam is working well for Coq, what would you think of including Flocq as an opam dependency instead of including it in the source tree? I would be up for doing the same thing for the parser's support library.

xavierleroy:
@jhjourdan: getting CompCert's dependencies through OPAM for Coq would be nice for the academic version of CompCert, but I'm not sure how it would work with the commercial version from AbsInt.

jhjourdan:
Yes, that would mean that we would have a release script for reconstituting a standalone source package for the commercial release, and then this commercial distribution would be much more different to the academic one than now.

ejgallego:
FYI, we are using this branch to test Coq trunk PRs before merging (example: https://travis-ci.org/coq/coq/builds/199146024 )

IMVHO Coq's new PR testing provides a small incentive for Compcert to become compatible with 8.6/8.7. 

We already found bugs revealed by 8.7 PR (race condition on extraction output !) that would have been quite painful to debug after merging.

xavierleroy:
All right, I imported this PR into branch `coq-8.6` of the AbsInt/CompCert repository, fixed the ARM and PowerPC code generators, updated Flocq to the latest release, and silenced the warnings using parts of pull request #150 .  Let me sleep on it and synchronize with AbsInt before merging to master.

PR #158, opened 2017-01-16
Option pipe

bschommer:
Adds the option -pipe which specifies that CompCert should uses pipes between the different process, if possible, instead of temporary files.

The Implementation works by introducing `open_process_{in|out}` like functionality for the preprocessor and the assembler as well as matching closing functions that either call the external process and use files for the input/output or start them in parallel and use pipes for input/output.

Furthermore the functionality for file names etc. is moved into the new module `File` and input files now have their own type `input_file` which stores the file with its suffix to allow easier manipulation of file suffices.


bschommer:
I will rebase this work in a new PR if #167, #168 and #169 are merged.

PR #159, opened 2017-01-20
Implement offsetof via builtin

bschommer:
The implementation of offsetof as macro in the form
((size_t) &((ty*) NULL)->member) has the problem that it cannot be
used everywhere were an integer constant expression is allowed,
for example in initiliazers of global variables and there is also
no check for the case that member is of bitifield type.

The new implementation adds a builtin function which is evaluated to
an integer constant during elaboration.

### DIFF
--- a/cparser/Cutil.ml
+++ b/cparser/Cutil.ml
@@ -533,6 +533,43 @@ let sizeof_struct env members =
       end
   in sizeof_rec 0 members
 
+(* Compute the offset of a struct member *)
+let offsetof env ty fields =
+  let align_field ofs ty =
+    let a =  match alignof env ty with
+      | Some a -> a
+      | None -> assert false in
+    align ofs a in
+  let rec offsetof_rec ofs field rest = function
+    | [] -> ofs
+    | m :: rem as ml ->
+      if m.fld_name = field.fld_name then begin
+        match rest with
+        | [] -> align_field ofs field.fld_typ
+        | _ -> lookup_field ofs field.fld_typ rest
+      end else if m.fld_bitfield = None then begin
+        match alignof env m.fld_typ, sizeof env m.fld_typ with
+        | Some a, Some s -> offsetof_rec (align ofs a + s) field rest rem
+        | _, _ -> assert false (* should never happen *)
+      end else begin
+        let (s, a, ml') = pack_bitfields ml in
+        let ofs = align ofs a + s in
+        offsetof_rec ofs field rest ml'
+      end
+  and lookup_field ofs ty = function
+    | [] -> align_field ofs ty
+    | fld::rest ->
+      begin match unroll env ty with
+        | TStruct (id,_) ->
+          let str = Env.find_struct env id in
+          offsetof_rec ofs fld rest str.ci_members
+        | TUnion (id,_) ->
+          lookup_field ofs fld.fld_typ rest
+        | _ -> assert false
+      end
+  in
+  lookup_field 0  ty (List.rev fields)
+
 (* Simplified version to compute offsets on structs without bitfields *)
 let struct_layout env members =
   let rec struct_layout_rec mem ofs = function
--- a/cparser/Cutil.ml
+++ b/cparser/Cutil.ml
@@ -534,26 +534,23 @@ let sizeof_struct env members =
   in sizeof_rec 0 members
 
 (* Compute the offset of a struct member *)
-let offsetof env ty fields =
+let offsetof env ty field =
   let rec sub acc name = function
     | [] -> List.rev acc
     | m::rem -> if m.fld_name = name then
         List.rev acc
       else
         sub (m::acc) name rem in
-  let offset (ofs,ty) field =
-    match unroll env ty with
+  match unroll env ty with
     | TStruct (id,_) ->
       let str = Env.find_struct env id in
       let pre = sub [] field.fld_name str.ci_members in
       begin match sizeof_struct env pre ,alignof env field.fld_typ with
       | Some s, Some a ->
-        (ofs + align s a),field.fld_typ
+        align s a
       | _ -> assert false end
-    | _ -> ofs,field.fld_typ
-  in
-  let fields = List.rev fields in
-  fst (List.fold_left offset (0,ty) fields)
+    | TUnion _ -> 0
+    | _ -> assert false
 
 (* Simplified version to compute offsets on structs without bitfields *)
 let struct_layout env members =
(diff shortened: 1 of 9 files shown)


xavierleroy:
I agree this is a big improvement over the current, not-really-defined-behavior implementation.  Something looks weird in `Cutil.offsetof`, though, see code review.

cparser/Cutil.ml
```diff
@@ -533,6 +533,28 @@ let sizeof_struct env members =
...
+      else
+        sub (m::acc) name rem in
+  let offset (ofs,ty) field =
+    match unroll env ty with
+    | TStruct (id,_) ->
+      let str = Env.find_struct env id in
+      let pre = sub [] field.fld_name str.ci_members in
+      begin match sizeof_struct env pre ,alignof env field.fld_typ with
+      | Some s, Some a ->
+        (ofs + align s a),field.fld_typ
```
xavierleroy: Skipping over a field of size `s` and alignment `a` is achieved by `align ofs a + s`, not `ofs + align s a`.

cparser/Cutil.ml
```diff
@@ -533,6 +533,28 @@ let sizeof_struct env members =
...
+let offsetof env ty fields =
+  let rec sub acc name = function
+    | [] -> List.rev acc
+    | m::rem -> if m.fld_name = name then
+        List.rev acc
+      else
+        sub (m::acc) name rem in
+  let offset (ofs,ty) field =
+    match unroll env ty with
+    | TStruct (id,_) ->
```
xavierleroy: Why a `TStruct` case and not a `TUnion` case?

cparser/Cutil.ml
```diff
@@ -533,6 +533,28 @@ let sizeof_struct env members =
...
+      let str = Env.find_struct env id in
+      let pre = sub [] field.fld_name str.ci_members in
+      begin match sizeof_struct env pre ,alignof env field.fld_typ with
+      | Some s, Some a ->
+        (ofs + align s a),field.fld_typ
+      | _ -> assert false end
```
xavierleroy: This is assuming that all involved types were checked to be complete types before.  It's probably the case.

bschommer: This is checked before in Elab.ml

cparser/Cutil.ml
bschommer: `ofs` is just a bad name here, it is just an accumulator passed to the this function. The function calculates the offset of the fields on the path and adds them up. So for
    
    struct {
        ...
        struct c{
            int b;
        };
    } a
    offsetof(struct a,c);

it would calculate `offsetof(struct a,<anon>_0) + offsetof(struct b,c)`.

cparser/Cutil.ml
bschommer: Because `offsetof(union a,b)` is always 0.

bschommer:
I will update this a PR with more comments and extend the implementation to also allow `offsetof(struct s,m1.m2)`

cparser/Cutil.ml
xavierleroy: That's a good reason.  But I would prefer to see 3 cases: TStruct, TUnion, and a catch-all `_ -> assert false`

cparser/Cutil.ml
xavierleroy: Your code would be clearer if you were to define a function `offsetof1` returning the offset and type of a single field in a struct or union type, then iterate `offsetof1`.  Then we can argue whether `offsetof1` is correct or not...

bschommer: I started implementing a new simpler version today which does exactly this. One of the problems of this implementation is that it can not handle array designators. But at the moment it is not ready to be committed.

cparser/Cutil.ml
```diff
@@ -533,6 +533,25 @@ let sizeof_struct env members =
...
+    | [] -> List.rev acc
+    | m::rem -> if m.fld_name = name then
+        List.rev acc
+      else
+        sub (m::acc) name rem in
+  match unroll env ty with
+    | TStruct (id,_) ->
+      let str = Env.find_struct env id in
+      let pre = sub [] field.fld_name str.ci_members in
+      begin match sizeof_struct env pre ,alignof env field.fld_typ with
```
xavierleroy: Spacing: comma-space instead of space-comma

cparser/Elab.ml
```diff
@@ -1642,6 +1642,46 @@ let elab_expr vararg loc env a =
...
+        let off,ty = offset_of_list 0 env ty flds in
+        env,off_accu + off,ty
+      | ATINDEX_INIT e,TArray (sub_ty,b,_) ->
+        let e,env = elab env e in
+        let e =
+          begin match Ceval.integer_expr env e,b with
+          | None,_ ->
+            error  "array element designator for is not an integer constant expression"
+          | Some n,Some b -> if n >= b then
+              error "array index %Ld exceeds array bounds" n;
```
xavierleroy: Neither gcc nor clang do an error in this case, not even a warning...  Note that the case `n` = `b` could even make sense as a way to get the offset of the point "one past" the last element.  I'd suggest to either do nothing, or do a warning if `n` > `b`.

cparser/Elab.ml
```diff
@@ -1642,6 +1642,46 @@ let elab_expr vararg loc env a =
...
+            n
+          | Some n,None -> assert false
+          end in
+        let size = match sizeof env sub_ty with
+          | None -> assert false (* We expect only complete types *)
+          | Some s -> s in
+        env,off_accu + size * (Int64.to_int e),sub_ty
```
xavierleroy: I am nervous about possible arithmetic overflow in these operations.  As a first step towards more resilience, `cautious_mul` from module Cutil could be used for the multiplication.

cparser/Elab.ml
```diff
@@ -1642,6 +1642,46 @@ let elab_expr vararg loc env a =
...
+      | ATINDEX_INIT _,_ -> error "subscripted value is not an array" in
+    let env,offset,_ = List.fold_left offset_of_member (env,0,ty) mem in
+    let offsetof_const = EConst (CInt(Int64.of_int offset,size_t_ikind (),"")) in
+    { edesc = offsetof_const; etyp = TInt(size_t_ikind(), []) },env
```
xavierleroy: While unlikely, it is possible for the value of `offset` to overflow the range of type `size_t`.  As a first step, using `Ceval.normalize_int` would ensure that a value of type `size_t` is produced.

cparser/Env.ml
```diff
@@ -220,6 +220,8 @@ let find_union_member env (id, m) =
   with Not_found ->
     raise(Error(No_member(id.name, "union", m)))
 
+
+
```
xavierleroy: Please remove those two blank lines.

bschommer:
The comments now should  be handled.

xavierleroy:
Looks good to me!  Merging right now.

PR #160, opened 2017-01-31
Revised elaboration of attributes

xavierleroy:
[This is work in progress and this pull request is intended for discussion.  Do not merge yet. ]

The treatment of attributes in the current CompCert is often surprising.  For example,
```
   attribute(XXX) char * x;
```
is parsed as "x is a pointer to a (char modified by attribute XXX)", while for most attributes (e.g. section attributes) the expected meaning is "x, modified by attribute XXX, has type pointer to char".

CompCert's current treatment comes from the fact that attributes are processed very much like the standard type modifiers `const` and `volatile`, i.e.
```
   const char * x;
```
is really "x is a pointer to a const char", not "x is a const pointer to char".

This experiment introduces a distinction between type-related attributes (which include the standard modifiers `const` and `volatile`) and other attributes.  The other, non-type-related attributes are "floated up" during elaboration so that they apply to the variable or function being declared or defined.  In the examples above,
```
   attribute(XXX) char * x;    // "attribute(XXX)" applies to "x"
   const char * x;             // "const" applies to "char"
```
This may be a step in the right direction but is not the final story.  In particular, the `packed` attribute is special-cased when applied to `struct`, like it was before, and future attributes concerning calling conventions would need to be floated up to function types but not higher than that.

### DIFF
--- a/cfrontend/C2C.ml
+++ b/cfrontend/C2C.ml
@@ -267,6 +267,19 @@ let builtins =
   { typedefs = builtins_generic.typedefs @ CBuiltins.builtins.typedefs;
     functions = builtins_generic.functions @ CBuiltins.builtins.functions }
 
+(** ** The known attributes *)
+
+let attributes = [
+  (* type-related *)
+  ("aligned", Cutil.Attr_type);
+  (* struct-related *)
+  ("packed", Cutil.Attr_struct);
+  (* function-related (currently none) *)
+  (* name-related *)
+  ("section", Cutil.Attr_name)
+]
+  
+
 (** ** Functions used to handle string literals *)
 
 let stringNum = ref 0   (* number of next global for string literals *)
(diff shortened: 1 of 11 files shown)


xavierleroy:
As reported by @bschommer something was wrong in the elaboration of array types. [be2192f] fixes that and [47c82de] is cosmetic.

xavierleroy:
Commit [3babd25] refactors the handling of custom attributes.  A table associates a "class" to each custom attribute, one among:
- `Attr_type`: the attribute is to be attached to the nearest type just like the qualifiers `const` and `volatile`.  Example: `aligned`.
- `Attr_struct`: the attribute is attached to a struct, union or enum.  Example: `packed`.
- `Attr_function`: the attribute is "floated up" and attached to the nearest enclosing function type.  Example: none currently, but `noreturn` or `fastcall` would be like this.
- `Attr_name`: the attribute is "floated up" all the way to the top declaration or definition of a variable or function.  Example: `section`.

The `Attr_function` behavior is new and will be exercised in later commits.  The other three behaviors are like before but implemented differently.

PR #162, opened 2017-02-07
Improved warnings related to function returns

xavierleroy:
This pull request introduces a control-flow static analysis over C abstract syntax (file cparser/Cflow.ml) and uses it to
- warn for non-void functions that can return by reaching the end of the function body;
- warn more precisely for `_Noreturn` functions that can return;
- avoid introducing the `return 0;` in `main` functions that always return already (cosmetic).

The static analysis is conservative in the sense that it makes no assumptions about the values of expressions used in conditional statements, unless the expressions are compile-time constants.  For example,
```
    if (x < 0) return -1;
```
is analyzed as possibly returning or possibly falling through.  In contrast,
```
   while (1) { ... }
```
is analyzed as never falling through.

### DIFF
--- a/cparser/Cflow.ml
+++ b/cparser/Cflow.ml
@@ -94,14 +94,16 @@ let catch_break (s: flow) : flow = fun i is ->
   let o = s i is in
   if can _Break o then add _Fallthrough (remove _Break o) else o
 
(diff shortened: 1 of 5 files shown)


bschommer:
The pull request looks good to me.

PR #163, opened 2017-02-07
New module to place checks on elaborated C code

bschommer:
The new module adds a function which is called during parse after all C transformation have taken place for adding additional checks. Currently only unknown attribute are checked.

### DIFF
--- a/cparser/Cerrors.ml
+++ b/cparser/Cerrors.ml
@@ -88,6 +88,7 @@ type warning_type =
   | Unknown_pragmas
   | CompCert_conformance
   | Inline_asm_sdump
+  | Unused_variable
 
 (* List of active warnings *)
 let active_warnings: warning_type list ref = ref [
@@ -135,6 +136,7 @@ let string_of_warning = function
   | Unknown_pragmas -> "unknown-pragmas"
   | CompCert_conformance -> "compcert-conformance"
   | Inline_asm_sdump -> "inline-asm-sdump"
+  | Unused_variable -> "unused-variable"
 
 (* Activate the given warning *)
 let activate_warning w () =
@@ -179,6 +181,7 @@ let wall () =
     Unknown_pragmas;
     CompCert_conformance;
     Inline_asm_sdump;
+    Unused_variable
   ]
 
 let wnothing () =
@@ -207,6 +210,7 @@ let werror () =
     Unknown_pragmas;
     CompCert_conformance;
     Inline_asm_sdump;
+    Unused_variable;
   ]
 
 (* Generate the warning key for the message *)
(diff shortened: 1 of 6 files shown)


bschommer:
I will rework the code a little.

xavierleroy:
Concerning the checking of attributes:
* Globally this looks very good to me.  
* I'm curious to know why the check is done so late (after all post-elaboration transformations) and not just after elaboration, like the unused variable analysis.  At any rate, the StructReturn transformation adds `__structreturn` attributes on some functions, and this attribute is classified as unknown because we don't want to expose it to the programmer, so false alarms will be reported on this attribute if the analysis is done after the StructReturn transformation.
* Besides function and variable declarations, I would also check struct, union and enum declarations.  They, too, contain attributes.
* Checking fo unknown attributes is good; checking for known but misplaced attributes could be even better!  For example, in the following, `packed` and `noreturn` are known but are applied to a variable declaration, where they don't make sense:
```
   attribute((packed,noreturn)) int x;
```

xavierleroy:
Concerning the unused variable analysis:
* Besides function-local variables, it would be nice to check for unused global variables and functions of the `static` kind (but not of the `static inline` kind!) and perhaps of the `extern` kind too.
* Owing to the use of `Identmap.iter`, if there are several unused variables, they will be reported in alphabetical order of variable name, not in the order in which they occur in the program.
* There is a different implementation that would avoid the problem above and perhaps be simpler overall.  A first pass through the program computes the set of identifiers used.  A second pass goes over each declaration in program order and warns immediately (at declaration point) if the declared name is unused.
* It would be a nice touch to have and honor GCC's `unused` attribute on declarations.  Macro-generated C code sometimes contains unused variable declarations, and it is good to be able to silence the warning on these particular declarations.

bschommer:
Concerning the checking of attributes:

>I'm curious to know why the check is done so late (after all post-elaboration transformations) and not just after elaboration, like the unused variable analysis. At any rate, the StructReturn transformation adds __structreturn attributes on some functions, and this attribute is classified as unknown because we don't want to expose it to the programmer, so false alarms will be reported on this attribute if the analysis is done after the StructReturn transformation.

It actually can be performed earlier.

>Checking fo unknown attributes is good; checking for known but misplaced attributes could be even better! For example, in the following, packed and noreturn are known but are applied to a variable declaration, where they don't make sense:

I agree actually gcc has a even a different warning for this and I will add it.

Concerning the unused variable analysis:

>Besides function-local variables, it would be nice to check for unused global variables and functions of the static kind (but not of the static inline kind!) and perhaps of the extern kind too.

Can be done. We also can warn about unused labels.

>There is a different implementation that would avoid the problem above and perhaps be simpler overall. A first pass through the program computes the set of identifiers used. A second pass goes over each declaration in program order and warns immediately (at declaration point) if the declared name is unused.

I wanted to traverse the code only once but it makes more sense to perform it as described.

>It would be a nice touch to have and honor GCC's unused attribute on declarations. Macro-generated C code sometimes contains unused variable declarations, and it is good to be able to silence the warning on these particular declarations.

 I will add this.

bschommer:
I rebased the PR since it needs some work.

PR #167, opened 2017-02-13
Introduced configuration variable for gnu systems.

bschommer:
The variable gnu_toolchain is true if a gnu toolchain is used and
false in all other cases. The variable avoids the explicit test
whether the system string is diab and should be easier to change.

Separated changes from #158 for easier review.


xavierleroy:
Fine with me.  Merging right now...

PR #168, opened 2017-02-13
Handle dcompcertc and dparsedc like all dump opts

bschommer:
The dcompcert and dparsedc option now also set a variable for the
dump file instead of being handled by looking at the clflag.

Another part of PR #158 .


bschommer:
Wrong place for the set calls, they need to be set earlier.

PR #169, opened 2017-02-13
Moved file functions in own file.

bschommer:
The file functions are moved into an own library file. The
temp_file function now registers a delete function at exit. This
avoids the double check for temp files in driver.

Currently unused is the input_file type and functions since most
functions using this are affected by the other changes in PR #158
and changing them would result in doubled work.

Another prequisite for the option pipe PR #158

### DIFF
--- a/driver/Assembler.ml
+++ b/driver/Assembler.ml
@@ -26,7 +26,7 @@ let assemble ifile ofile =
   ] in
   let exc = command cmd in
   if exc <> 0 then begin
-    safe_remove ofile;
+    File.safe_remove ofile;
     command_error "assembler" exc;
     exit 2
   end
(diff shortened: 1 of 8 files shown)


xavierleroy:
Just had a quick look, need to read again.  

+1 for the temporary files with automatic removal at exit, that simplifies things nicely.

Less sure about the other changes.  Probably I should look at #158 to see where you're going with the `input_file` stuff.

bschommer:
The input file is used to simplify the whole handling of suffix cutting etc. and it also includes the check whether the file exists or not.

xavierleroy:
That's one thing I wanted to mention in a future code review: checking that a file exists at time T gives no guarantee that the file can be read later, i.e. that open_in{_bin} will succeed.  (File could have been removed, or could have the wrong permissions, etc.)

bschommer:
The main reason for checking this in CompCert is to avoid for passing non-existing files to the external tools, we could extend the check to test whether the file has the right permissions.

bschommer:
I will rebase this and add a check for file permissions etc.

PR #170, opened 2017-02-14
Removed Cminor input

bschommer:
As the title indicates this PR removes the CMinor input.


xavierleroy:
There are also some tests to remove, but I'll do that later.

xavierleroy:
Tests removed in commit [master 4ac4530]

PR #172, opened 2017-02-16
Added handling for noreturn std functions.

bschommer:
The C11 standard declares exit,abort,_Exit,quick_exit and
thrd_exit as _Noreturn however this is not included in older C
libs and leads to false negatives in reporting _Noreturn and
return type warnings. This can be avoided by enhancing the
noreturn check of the Cflow analysis to also test if one of the
above functions is called.

### DIFF
--- a/cparser/Cflow.ml
+++ b/cparser/Cflow.ml
@@ -24,7 +24,7 @@ module StringSet = Set.Make(String)
 
 (* Functions declared noreturn by the standard *)
 let std_noreturn_functions =
-   ["longjmp";"exit";"abort";"_Exit";"quick_exit";"thrd_exit"]
+   ["longjmp";"exit";"_exit";"abort";"_Exit";"quick_exit";"thrd_exit"]
 
 (* Statements are abstracted as "flow transformers":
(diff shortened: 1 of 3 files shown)


xavierleroy:
Ok with me.  Having function names hard-wired in the compiler is certainly less nice than having the functions properly declared in the proper headers, but for the purposes of warnings it's OK.

I never saw _Exit,quick_exit and thrd_exit used in actual code, but there are several other, commonly-used  functions that don't return: _exit, longjmp, siglongjmp.  Where to draw the line?  I would add longjmp (because it is part of the ISO C standard) and _exit (because it is very common).

bschommer:
The functions are just the ones declared with _Noreturn in C11. I overlook _Noreturn and will add it.

bschommer:
As for the header files: I think the problem is that either they have no idea about `_Noreturn` (diab's dcc uses C90 as default) or they hide the necessary attribute because we are a non gnu compiler.

xavierleroy:
Can we please have _exit as well?

xavierleroy:
Thanks a lot!  Merging right now.

PR #173, opened 2017-02-17
Moved file functions in own file. [Rebase of #169]

bschommer:
The file functions are moved into an own library file. The
temp_file function now registers a delete function at exit. This
avoids the double check for temp files in driver.

Currently unused is the input_file type and functions since most
functions using this are affected by the other changes in PR #158
and changing them would result in doubled work.


bschommer:
The `new_input_file` now also test whether the file can be read.

bschommer:
Mostly subsumed by other commits.

PR #175, opened 2017-03-08
Adapt proofs to future handling of literal constants in Coq.

silene:
This commit is mainly a squash of the relevant compatibility commits from Flocq's master. Most of the changes are meant to make the proofs oblivious to the way constants such as 0, 1, 2, and -1 are represented.

The main point of this commit is to keep CompCert's master as part of Coq's regression suite.


xavierleroy:
Thanks for those changes.  It is valuable to stay in sync with Flocq and to anticipate future Coq evolutions.  Merging right now.

PR #184, opened 2017-05-27
Adapt to a fix in zify

letouzey:
See Coq pull request #673 (and original bug #5336).
With the fixed version of zify, this proof in Inliningspec.v could actually be shortened to `intros. unfold shiftpos. now zify.`, but the proposed patch has the advantage of being compatible with both the released versions of Coq, and the coming ones.

### DIFF
--- a/backend/Inliningspec.v
+++ b/backend/Inliningspec.v
@@ -72,7 +72,7 @@ Qed.
 
 Lemma shiftpos_eq: forall x y, Zpos (shiftpos x y) = (Zpos x + Zpos y) - 1.
 Proof.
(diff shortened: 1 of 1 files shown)


ejgallego:
@letouzey you can link directly to the Coq PR using `coq/coq#NUM`, as in coq/coq#673. We have nothing for buzgilla thou, so https://coq.inria.fr/bugs/show_bug.cgi?id=5336 has to be used.

xavierleroy:
Thanks for the patch, and glad to learn that "zify" is getting even better.  Merged just now.

PR #185, opened 2017-06-06
Do not rely on the hints of BigNumPrelude

letouzey:
BigNumPrelude will soon leave Coq stdlib with the rest of the bignums library (apart from Int31 files) to become a separate package, see [PR#498](https://github.com/coq/coq/pull/498).

With this (very minor) patch, Compcert compiles with or without the hints declared in BigNumPrelude.

Best,
Pierre L.

### DIFF
--- a/cparser/validator/Alphabet.v
+++ b/cparser/validator/Alphabet.v
@@ -249,7 +249,7 @@ erewrite <- IHn, <- Zabs_nat_Zsucc in H |- *; eauto; try omega.
 rewrite phi_incr, Zmod_small; intuition; try omega.
 apply inj_lt in H.
 pose proof Zle_le_succ.
-do 2 rewrite inj_Zabs_nat, Zabs_eq in H; eauto.
+do 2 rewrite inj_Zabs_nat, Zabs_eq in H; now eauto.
 Qed.
 
 (** Previous class instances for [option A] **)


xavierleroy:
Looks fine to me.  @jhjourdan : any objections?

xavierleroy:
The diff is incredibly small, that's black magic to me, but I take @letouzey 's work that it does the right thing!

jhjourdan:
I don't have any objection. I have no idea why this works, though... I though `now` were just a synonymous to `; fail`.

letouzey:
@jhjourdan : in fact, `now` stands for ` ; easy`, and `easy` is a kind of vade-vecum not yet documented properly ... but roughly including `trivial`, `reflexivity`, `discriminate`, and a few other fast goodies. And indeed, `easy` either solve the current goal or  `fail`.

About the proposed change : with the hints inside `BigNumPrelude`, the final  `eauto` of this proof used to be slightly stronger, and solved a last `0 <= 2^32` goal. Without these hints, this  `0 <= 2^32` remains, which reduces to `Lt <> Gt`, solvable by a `discriminate` (and hence by `now`).

xavierleroy:
Thanks for the tips about "easy", it's always good to learn from the masters.

Merging right now.

PR #188, opened 2017-06-30
Add a -ignore-coq-version flag to configure.

maximedenes:
The version check in configure is very strict (it does not allow to use
a different minor version of Coq). In some use cases, like continuous
testing on the Coq side, it is useful to bypass the check.

Note that the version of Coq is still explicitly shown as unsupported.

### DIFF
--- a/configure
+++ b/configure
@@ -21,6 +21,7 @@ has_runtime_lib=true
 has_standard_headers=true
 clightgen=false
 responsefile="gnu"
+ignore_coq_version=false
 
 usage='Usage: ./configure [options] target
 
@@ -69,6 +70,7 @@ Options:
   -no-runtime-lib      Do not compile nor install the runtime support library
   -no-standard-headers Do not install nor use the standard .h headers
   -clightgen           Also compile the clightgen tool
+  -ignore-coq-version  Do not check the version of Coq
 '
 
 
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -69,6 +70,7 @@ Options:
   -no-runtime-lib      Do not compile nor install the runtime support library
   -no-standard-headers Do not install nor use the standard .h headers
   -clightgen           Also compile the clightgen tool
+  -ignore-coq-version  Do not check the version of Coq
```
ejgallego: @maximedenes maybe a different wording could be used here? "Try to compile even with unsupported Coq versions."

maximedenes: Sure, feel free to submit a PR on my branch, I'm very busy. Thanks!

maximedenes:
@xavierleroy Does this look acceptable to you? We can tweak it if required, but it would be useful to improve the continuous testing of CompCert on the Coq repo.

configure
ejgallego: Waiting on feedback from Comcert folks. I suggest you add the 8.7 to Coq's 8.6 travis.

xavierleroy:
This feature is fine, thanks for the PR.  Merged manually in commit 47700fa .  Why the manual merge?  First I tweaked the usage message as suggested by @ejgallego.  Then I wasn't sure how portable `if ! ` is so I wrote an if..then..else instead.

maximedenes:
Thanks, Xavier!

PR #191, opened 2017-07-20
Ensure FunInd or Recdef is imported if functional induction is used

sigurdschneider:
Coq 8.7 does not load FunInd in prelude anymore, so importing FunInd becomes necessary in a couple of places. 

In Coq 8.7, `Recdef` exports `FunInd`, hence if `Recdef` is imported, `FunInd` does not have to be required separately.

With this change, CompCert compiles with Coq 8.7+alpha (coq/coq@d074e889b3cdfe8c292d3c52a4ed005789384fc0). This change should still be compatible with Coq 8.6. I did one build each for ppc, arm, and x86 with Coq 8.6.1.

An **alternative to this patch** is to just add `Require Export FunInd` in `Coqlib.v`. If that is prefered, feel free to close this PR.

### DIFF
--- a/arm/CombineOpproof.v
+++ b/arm/CombineOpproof.v
@@ -13,6 +13,7 @@
 (** Recognition of combined operations, addressing modes and conditions
   during the [CSE] phase. *)
 
+Require Import FunInd.
 Require Import Coqlib.
 Require Import AST.
 Require Import Integers.
--- a/backend/Allocproof.v
+++ b/backend/Allocproof.v
@@ -13,6 +13,7 @@
 (** Correctness proof for the [Allocation] pass (validated translation from
   RTL to LTL). *)
 
+Require Import FunInd.
 Require Import FSets.
 Require Import Coqlib Ordered Maps Errors Integers Floats.
 Require Import AST Linking Lattice Kildall.
--- a/backend/Deadcodeproof.v
+++ b/backend/Deadcodeproof.v
@@ -12,6 +12,7 @@
 
 (** Elimination of unneeded computations over RTL: correctness proof. *)
 
+Require Import FunInd.
 Require Import Coqlib Maps Errors Integers Floats Lattice Kildall.
 Require Import AST Linking.
 Require Import Values Memory Globalenvs Events Smallstep.
--- a/backend/Selectionproof.v
+++ b/backend/Selectionproof.v
@@ -12,6 +12,7 @@
 
 (** Correctness of instruction selection *)
 
+Require Import FunInd.
 Require Import Coqlib Maps.
 Require Import AST Linking Errors Integers Values Memory Events Globalenvs Smallstep.
 Require Import Switch Cminor Op CminorSel.
--- a/backend/ValueAnalysis.v
+++ b/backend/ValueAnalysis.v
@@ -10,6 +10,7 @@
 (*                                                                     *)
 (* *********************************************************************)
 
+Require Import FunInd.
 Require Import Coqlib Maps Integers Floats Lattice Kildall.
 Require Import Compopts AST Linking.
 Require Import Values Memory Globalenvs Events.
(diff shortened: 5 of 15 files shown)


m-schmidt:
Thanks, we will have a look at it.

xavierleroy:
I wasn't sure this Function-related change in Coq 8.7 was intentional, but yes, it is: 
https://github.com/coq/coq/commit/27c8e30fad95d887b698b0e3fa563644c293b033

Zimmi48:
Indeed, this was an intentional change (see coq/coq#220).
Note that because of this change, the version of CompCert that is currently tested for compatibility with Coq is not the official one, but the less_init_plugins branch of @letouzey's fork, which has this additional commit: https://github.com/letouzey/CompCert/commit/fc58591af0a9f639f77acc254fbcadd0cf0da74f adding `Require Export FunInd` in `Coqlib.v` as mentioned as a possible fix by @sigurdschneider.
Note also that the `FunInd.v` file was introduced at the time of the PR in 8.7 and also in 8.6.1 for backward compatibility, but compatibility with 8.6(.0) is not maintained.
If compatibility with 8.6(.0) is important, there is the alternative possibility of importing `Recdef` as @sigurdschneider also mentioned because this file existed in 8.6(.0) already but this imports more stuff at the same time.

xavierleroy:
> Note also that the FunInd.v file was introduced at the time of the PR in 8.7 and also in 8.6.1 for backward compatibility, but compatibility with 8.6(.0) is not maintained.

This is not nice.  It would have been more professional to introduce FunInd.v, wait one full 8.7 release cycle, then make FunInd.v mandatory in Coq 8.8.  I mean, it's not like this change is fixing a serious bug in Coq, it's just twiddling with stuff that already worked.

At any rate, the next CompCert release (3.1, planned end August 2017) must remain compatible with 8.6, so I won't merge these changes until CompCert 3.1 is released.

Zimmi48:
Well, there was a compatibility effort made here. The PR author didn't even propose initially any backward compatible fix. This is also an approach that was used before when Ltac moved to a plugin (a dummy Ltac plugin was created in 8.5pl3, even though only plugin writers noticed).

That being said, I do not wish to take any responsibility in this particular compatibility-breaking change so I won't try to defend it more.

I'll try to report your view on this next time something similar happens.

maximedenes:
> If compatibility with 8.6(.0) is important, there is the alternative possibility of importing Recdef as @sigurdschneider also mentioned because this file existed in 8.6(.0) already but this imports more stuff at the same time.

Yes, but `Recdef` defines a toplevel constant called `max`, so expect some problems there.

So I think we could have been more professional indeed. From the outside, the benefit/cost ratio of this change is not great. Apologies for that.

Hopefully this kind of things should not happen anymore. For this particular instance, I had to push for at least *some* backward compatibility, but our long release cycle made it hard to be stricter on the deprecation process. With our new 6-month cycle, things will hopefully be easier.

xavierleroy:
Apparently I came across as very grumpy, while I'm just a little grumpy :-)  

With the introduction of Funind.v in 8.6.1, I appreciate the attempt at anticipating an incompatible change in 8.7.  I'm just pointing out the obvious: it will still be impossible to be compatible with 8.6, 8.6.1 and 8.7 at the same time.  (And that minor releases such as 8.6.1 are the wrong place for adding features of any kind, even of the "future compatibility" kind, but that's my grumpiness showing again.)

Concerning CompCert, this has no bad consequences.  It's just that I'll integrate this PR when CompCert switches to Coq 8.7, but not before.

xavierleroy:
Now that Coq 8.7.0 is out, it's time to merge this PR and switch the CompCert master to Coq 8.7.0 OR 8.6.1.  Thanks for having submitted this PR.

sigurdschneider:
Thanks!

PR #192, opened 2017-08-02
More C compliant Clight printer

fajb:
The pull request solves minor issues of the Clight printer.
With the modifications, more Clight programs are compliant C programs.

Best,

### DIFF
--- a/cfrontend/PrintClight.ml
+++ b/cfrontend/PrintClight.ml
@@ -79,9 +79,9 @@ let rec expr p (prec, e) =
   | Econst_int(n, _) ->
       fprintf p "%ld" (camlint_of_coqint n)
   | Econst_float(f, _) ->
-      fprintf p "%.15F" (camlfloat_of_coqfloat f)
+     fprintf p "%.18g" (camlfloat_of_coqfloat f)
   | Econst_single(f, _) ->
-      fprintf p "%.15Ff" (camlfloat_of_coqfloat32 f)
+     fprintf p "%.18g" (camlfloat_of_coqfloat32 f)
   | Econst_long(n, Tlong(Unsigned, _)) ->
       fprintf p "%LuLLU" (camlint64_of_coqint n)
   | Econst_long(n, _) ->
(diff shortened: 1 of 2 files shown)


cfrontend/PrintClight.ml
```diff
@@ -79,9 +79,9 @@ let rec expr p (prec, e) =
   | Econst_int(n, _) ->
       fprintf p "%ld" (camlint_of_coqint n)
   | Econst_float(f, _) ->
-      fprintf p "%.15F" (camlfloat_of_coqfloat f)
+     fprintf p "%.18g" (camlfloat_of_coqfloat f)
```
xavierleroy: Why the change from `.15F` to `.18g` ?

cfrontend/PrintClight.ml
```diff
@@ -79,9 +79,9 @@ let rec expr p (prec, e) =
...
   | Econst_single(f, _) ->
-      fprintf p "%.15Ff" (camlfloat_of_coqfloat32 f)
+     fprintf p "%.18g" (camlfloat_of_coqfloat32 f)
```
xavierleroy: Here you must leave a `f` or `F` suffix to indicate `float` type instead of the default `double` type.

cfrontend/PrintClight.ml
```diff
@@ -116,22 +116,28 @@ let rec print_expr_list p (first, rl) =
 
 (* Statements *)
 
+(* Receiver of call *)
+let print_rec p e =
```
xavierleroy: The `_rec` suffix means "recursive" to me.  Also, you shoud use the existing `expr` printer with precedence 15. (See below.)

cfrontend/PrintClight.ml
```diff
@@ -116,22 +116,28 @@ let rec print_expr_list p (first, rl) =
...
-      fprintf p "/*skip*/;"
+      fprintf p "(void)0 /*skip*/;"
   | Sassign(e1, e2) ->
       fprintf p "@[<hv 2>%a =@ %a;@]" print_expr e1 print_expr e2
   | Sset(id, e2) ->
       fprintf p "@[<hv 2>%s =@ %a;@]" (temp_name id) print_expr e2
   | Scall(None, e1, el) ->
       fprintf p "@[<hv 2>%a@,(@[<hov 0>%a@]);@]"
-                print_expr e1
+                print_rec e1
```
xavierleroy: Here and below, just use `expr (15, e1)` instead of `print_rec e1`.

cfrontend/PrintClight.ml
```diff
@@ -266,19 +272,35 @@ let print_globdef p (id, gd) =
...
+let collect_internal_fct defs =
+  let add_decl (id,gd) l =
+    match gd with
+    | AST.Gfun (Internal f) -> (id,f)::l
+    |  _ -> l in
+  List.fold_right add_decl defs []
+
+let print_fundecl p (id,fct) =
+  fprintf p "%s;@ @ " (name_cdecl (extern_atom id) (Clight.type_of_function fct))
+
```
xavierleroy: There is no real need for building a list of what you're going to print, then print by iterating over this list.  Why not iterate over the whole list of definitions and choose what you're going to print?

cfrontend/PrintClight.ml
fajb: I suspect that %.15F is not precise enough -- I am using ocaml 4.05.0.
This can be seen by the following program:
#include <stdio.h>
#define PI 3.14159265358979323846
int main(){  printf("PI = %.15f\n",PI); }

The compiled C program and the dumped Clight program do not agree.
At Clight level, I get   printf(__stringlit_1, 3.14159265359);
And at runtime, PI = 3.141592653589793        vs     PI = 3.141592653590000

The problem vanishes with "%.18g" which is actually used by the x86 assembly printer.

cfrontend/PrintClight.ml
fajb: More elegant indeed.

cfrontend/PrintClight.ml
fajb: This could indeed be done on-the-fly.

xavierleroy:
This pull request doesn't work: no only the changes I requested are not done, but an unrelated commit (
3a17db3) was pushed on top.  I'm closing this pull request and will merge manually the relevant parts.

xavierleroy:
Manual merge in [master c0e121c], feel free to test and comment on if it doesn't do the right thing.

I see the Cminor printing changes were merged in [master 115825a].

Issue #199, opened 2017-08-16
CompCert as a library, building CompCert with -Q


andrew-appel:
When CompCert is used as a library imported into other Coq projects, it is a problem that CompCert is built with a loadpath that uses -R instead of -Q.  

The front end of CompCert, up to and including common/Clight.v and exportclight/Clightdefs.v, is used as a library imported into VST and perhaps other projects.   Now that VST is also used as a library imported into arbitrary user projects, we want to make sure that VST does not pollute the namespace of these other projects.  That means we build VST with -Q in the loadpath instead of -R.

However, the user of VST must also import the front end of CompCert, up to Clightdefs.  If that is built with -R, then all the names defined in (the front end of) CompCert get imported into the user's project.

I propose two solutions to this problem.
1.  Build CompCert with -Q instead of -R.   That is, change the COQINCLUDES in the Makefile to use -Q instead of -R.  This requires changing many "Require" statements in the .v files, but that should be fairly straightforward.

2.  Alternate hack, which would improve the situation from the current one, but is imperfect:
Build CompCert with -R, as it is now, but import CompCert into other projects using -Q.
This hack requires minor changes to Clightdefs.v and ExportClight.ml.
The problem with this hack is that the Module namespace is still polluted.

In either case, Clightdefs.v must be modified as follows:
* All the Require statements should use fully qualified paths.
* Do not "Require Export" in the Clightdefs.v  (that pollutes the namespace for anyone that has to import Clightdefs).  I recommend changing all the Export to Import, but this also requires that ExportClight.ml print one additional line after it prints "Require Import Clightdefs", that is: 
"Import Coqlib Integers Floats AST Ctypes Cop Clight."

The reason that solution 2 still pollutes the namespace (slightly) is that "Clight.Tpointer" is now visible to anyone who imports Clightdefs.v, instead of "compcert.common.Clight.Tpointer".  But I believe (if I understand correctly) that in solution 1, a user could Import compcert.exportclight.Clightdefs, and then the shortest path to Tpointer would be compcert.common.Clight.Tpointer.

andrew-appel:
There will be one issue to resolve with solution 1:  $(ARCH).  In the Makefile it's easy enough to make that a defined name, but a fully qualified Require statement will need a static name.

One solution that might work is:  omit $(ARCH) from RECDIRS, instead special-case it in the COQINCLUDES as,   -Q $(ARCH)  compcert.architecture.

[...]

PR #200, opened 2017-08-18
CompCert for Lib

bschommer:
Use -Q instead of -R for the include's in order to avoid polluting the global namespace as outlined in issue #199. Note that the -Q option is used for all modules with compcert except for flocq.

In order to test this one needs to apply the patch attached from the [menhir issue](https://gitlab.inria.fr/fpottier/menhir/issues/1). Thus this cannot be merged before the patch is applied to menhir or something similar is implemented.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,9 @@ DIRS=lib common $(ARCHDIRS) backend cfrontend driver debug\
   flocq/Core flocq/Prop flocq/Calc flocq/Appli exportclight \
   cparser cparser/validator
 
-RECDIRS=lib common $(ARCHDIRS) backend cfrontend driver flocq exportclight cparser
+RECDIRS=lib common $(ARCHDIRS) backend cfrontend driver exportclight cparser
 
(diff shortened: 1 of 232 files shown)


xavierleroy:
Would the following Coq syntax work here?
```
  From compcert Require Import Coqlib Integers Floats.
```

bschommer:
Yes that would work.

bschommer:
Any preferences? I would prefer the `From compcert Require Import ...`.

xavierleroy:
I, too, prefer `From compcert...` but I don't want 20 lines of
```
From compcert Require Import Foo.
From compcert Require Import Bar.
```
So, this is not just a pass of `sed` on all the Coq files.  We should seize the opportunity to group imports in ways that are less verbose and more semantically meaningful (something I started to do in release 3.0 for some files as I was modifying them).

bschommer:
Sounds like a good plane.

bschommer:
There is another implementation solution for the given problem.

PR #202, opened 2017-09-15
Strength reduction patterns for ARM mla instruction.

gergo-:
Extend the ARM constant folding pass to simplify mla (multiply-add) operations in cases like `0 * y + z`, `1 * y + z`, `x * y + 0`, etc.


xavierleroy:
Thanks for the patch.  Looks good to me.  Merging!

PR #203, opened 2017-09-15
Generalize load-immediate operations on ARM

gergo-:
On ARM, `movw` and `movt` can always be used to load immediates into general-purpose registers. This used to be only done in Thumb mode. It is simpler than the series of `orr` operations that used to be emitted in ARM mode.

Also, the ARMv6 flavor targeted by CompCert is in fact ARMv6T2 (with Thumb2).

### DIFF
--- a/arm/Archi.v
+++ b/arm/Archi.v
@@ -66,5 +66,6 @@ Global Opaque ptr64 big_endian splitlong
 Inductive abi_kind := Softfloat | Hardfloat.
 Parameter abi: abi_kind.
 
-(** Whether to generate [movw] and [movt] instructions. *)
-Parameter move_imm: bool.
+(** Whether instructions added with Thumb2 are supported. True for ARMv6T2
+  and above. *)
+Parameter thumb2_support: bool.
--- a/arm/Archi.v
+++ b/arm/Archi.v
@@ -65,3 +65,6 @@ Global Opaque ptr64 big_endian splitlong
 
 Inductive abi_kind := Softfloat | Hardfloat.
 Parameter abi: abi_kind.
+
+(** Whether to generate [movw] and [movt] instructions. *)
+Parameter move_imm: bool.
--- a/driver/Driver.ml
+++ b/driver/Driver.ml
@@ -205,7 +205,8 @@ let version_string =
   else
     "The CompCert C verified compiler, version "^ Version.version ^ "\n"
 
-let target_help = if Configuration.arch = "arm" then
+let target_help =
+  if Configuration.arch = "arm" && Configuration.model <> "armv6" then
 {|Target processor options:
   -mthumb        Use Thumb2 instruction encoding
   -marm          Use classic ARM instruction encoding
@@ -372,7 +373,8 @@ let cmdline_actions =
   Exact "-conf", Ignore; (* Ignore option since it is already handled *)
   Exact "-target", Ignore;] @ (* Ignore option since it is already handled *)
   (if Configuration.arch = "arm" then
-    if Configuration.model = "armv6" then [] (* Thumb needs ARMv6T2 or ARMv7 *)
+    if Configuration.model = "armv6" then
+      [ Exact "-marm", Ignore ] (* Thumb needs ARMv6T2 or ARMv7 *)
     else
       [ Exact "-mthumb", Set option_mthumb;
         Exact "-marm", Unset option_mthumb; ]
--- a/driver/Driver.ml
+++ b/driver/Driver.ml
@@ -372,8 +372,10 @@ let cmdline_actions =
   Exact "-conf", Ignore; (* Ignore option since it is already handled *)
   Exact "-target", Ignore;] @ (* Ignore option since it is already handled *)
   (if Configuration.arch = "arm" then
-    [ Exact "-mthumb", Set option_mthumb;
-      Exact "-marm", Unset option_mthumb; ]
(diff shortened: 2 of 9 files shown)


xavierleroy:
Indeed, the question is whether we want to support "plain" ARMv6 (before ARMv6T2) or not.  If the answer is "yes" we need to keep the old, ORR-based code.

> Also, the ARMv6 flavor targeted by CompCert is in fact ARMv6T2 (with Thumb2).

What makes you conclude this?  If the option `-thumb` is not given, doesn't CompCert generate "plain" ARMv6 (not T2) code?

gergo-:
> If the option -thumb is not given, doesn't CompCert generate "plain" ARMv6 (not T2) code?

Yes, it does. Sorry if I was unclear. My understanding is that if `-mthumb` is *not* specified, it doesn't matter whether the architecture is called `armv6` or `armv6t2`. I'm not aware of differences in ARM mode among these architectures (but I'm not an expert on this corner of the ecosystem). The test suite passes.

However, *if* CompCert is configured for `armv6`, and `-mthumb` is given, then it generates Thumb version 2, and the architecture *must* be called `armv6t2` in the assembly code.

So I think the check in the target printer could also be written as follows:
`       | "armv6" -> (if !Clflags.option_mthumb then "armv6t2" else "armv6")
`
and a similar guard could be added to runtime/arm/sysdeps.h. But I don't think it would add value.

gergo-:
> I'm not aware of differences in ARM mode among these architectures

Except, of course, exactly the very instructions being discussed here. Sorry, I'm getting confused by this topic. However:

> Indeed, the question is whether we want to support "plain" ARMv6 (before ARMv6T2) or not.

If the answer to that question is yes, compilers configured for `armv6` and `armv6t2` would have to behave differently: The `armv6` compiler would have to reject the `-mthumb` flag, because CompCert has no Thumb version 1 code generator.

gergo-:
Added a patch to separate the ARMv6 and ARMv6T2 cases. There are now the following cases:

- ARMv6. This is "plain" ARM without any version of Thumb. `-mthumb` and `-marm` flags are not supported, `movw` and `movt` are never generated.
- ARMv6T2. Supports Thumb (version 2). Supports and generates `movw` and `movt` in both ARM and Thumb modes.
- ARMv7 (various flavors). Supports Thumb. Supports and generates `movw` and `movt` in both ARM and Thumb modes.

The generation of `movw` and `movt` therefore now only depends on the configuration, not on command line flags. The choice is represented by the new `Archi.move_imm` parameter.

For the `Archi.move_imm = false` case (plain ARMv6), I added the old `orr` immedate generation back in.

Let me know if you think this is the way to go.

xavierleroy:
This looks very good to me, thanks.  A few nits in the code review that follows.

I don't know if there is commercial interest in "plain" ARMv6 (pre-Thumb-2) processors.  Some Raspberry Pis were plain ARMv6: I remember because I had to buy a more modern ARM platform to test the Thumb2 mode of CompCert.  At any rate, it cannot hurt to keep plain ARMv6 support in CompCert for the time being.

Just for the record: besides `movw` and `movt` ARMv6T2 introduces a few other instructions that are available both in ARM and Thumb2 encodings, esp. bit-field manipulation instructions `bfc`, `bfi`, `sbfx`, `ubfx`.

arm/Archi.v
```diff
@@ -65,3 +65,6 @@ Global Opaque ptr64 big_endian splitlong
 
 Inductive abi_kind := Softfloat | Hardfloat.
 Parameter abi: abi_kind.
+
+(** Whether to generate [movw] and [movt] instructions. *)
+Parameter move_imm: bool.
```
xavierleroy: I would prefer a different name and intuitive meaning for this parameter, one that means "plain ARMv6" versus "ARMv6T2 and above".  One reason being that there are other instructions that are not in "plain ARMv6" and that we may want to generate some day.  Some possible names:
* `thumb2_support` or `T2_support`
* `armv6_only` (with the opposite meaning: `true` for ARMv6, `false` for ARMv6T2 and up).

driver/Driver.ml
```diff
@@ -372,8 +372,10 @@ let cmdline_actions =
   Exact "-conf", Ignore; (* Ignore option since it is already handled *)
   Exact "-target", Ignore;] @ (* Ignore option since it is already handled *)
   (if Configuration.arch = "arm" then
-    [ Exact "-mthumb", Set option_mthumb;
-      Exact "-marm", Unset option_mthumb; ]
+    if Configuration.model = "armv6" then [] (* Thumb needs ARMv6T2 or ARMv7 *)
+    else
```
xavierleroy: Just to nit-pick: the `-marm` option should be available on ARMv6 as well, so that a user who absolutely wants ARM encoding can express this intention by putting `-marm`, and still be able to compile on ARMv6.

gergo-:
Thanks for the review, I added a new patch. Good point about `sbfx` in particular, which can now also be generated in ARM on models with Thumb2 support.

xavierleroy:
I forgot that CompCert is already using the `sbfx` instruction!  It's good to be able to use it in ARM mode as well as in Thumb mode.

The current code looks good to me.  Maybe @m-schmidt or @bschommer would like to have a quick look at it before we merge?

bschommer:
Looks good to us.

xavierleroy:
All right, I merge, then.

Issue #208, opened 2017-11-21
optimisation due to undefined behaviour


fajb:
Hello,

I have the following silly program undef.c  which compares a pointer and an integer.
Without optimisation, the program does nothing but prints "BAD" with optimisation.
To me, this is a case where constant propagation seems to be a bit too aggressive.
Statically, it should know that the condition &x != 0x601024 is undefined
 but prefers to return Some true and therefore predicts that the branch is always taken.
If the dataflow concludes that a condition is necessarily undef, it looks safer to not optimise the code.

#include <stdio.h>
#include <stdint.h>
int x = 5; // I know that x is allocated at address 0x601024

int main(){
  uintptr_t p = (uintptr_t) &x;
  if (p != (uintptr_t) 0x601024)
    printf("BAD %p",p);
}

Best,
--
Frédéric Besson
PS: I triggered the bug in *real* life when adding numeric assertions about pointer values that I knew should hold at runtime.

jeehoonkang:
> If the dataflow concludes that a condition is necessarily undef, it looks safer to not optimise the code.

> I triggered the bug in *real* life

In terms of the quality of implementation, maybe it'd be better for the compiler to emit a warning in this case.  However, I'd like not to characterize this translation as a bug of CompCert: it's correct in terms of both CompCert's formally defined C semantics and the ISO C standards.

xavierleroy:
@jeehoonkang is correct that this is not a bug properly speaking, but a quality of implementation issue.  

@fajb is correct that this kind of optimization of undefined behaviors is best avoided. Indeed, CompCert's value analysis and constant propagation already tries hard not to optimize on this basis.  I'll look into this particular example ASAP.

fajb:
I reported the issue (indeed not a bug) because  I am aware the design of the value analysis aims at avoiding those  situations.  I also find the issue intriguing: there is some insight about what should not happen but no formal statement...

For the case at hand,  a generic countermeasure may be something like:
Let av be the result of a sound transfer function. For any other sound abstract value av',
we have either av' <= av (the transfer function is not optimal)
                  or   av  <= av' (there is no sound incomparable value)
I think this would prevent the transfer function to make some arbitrary guess. 
For the undefined comparison of pointers, this would force the transfer function to return either bot or top.

Another related issue is that the optimisation strength is not monotonic with the result of the value analysis. In particular, if the analysis computes undef, CompCert wisely does not perform optimisations.
[...]

PR #209, opened 2017-11-22
Make value analysis of comparisons more conservative w.r.t. pointers

xavierleroy:
This addresses issue #208.  

Comparisons such as `(uintptr_t) &global == 0x1234` are undefined behavior in CompCert but their status in ISO C is unclear and they may occur in real-world code.  

This PR makes sure they are analyzed as "I don't know" (`Btop`) by the value analysis of the `cmpu` and `cmplu` operators.  However, if a known address is compared against the null pointer or the integer 0, we keep the previous analysis and predict that they are not equal.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,8 @@
 Code generation and optimization:
 - ARM in Thumb mode: simpler instruction sequence for branch through jump table.
+- Issue #208: make value analysis of comparisons more conservative for
+  dubious comparisons such as "(uintptr_t) &global == 0x1234" which are
+  undefined behavior in CompCert.
 
 Usability:
 - Resurrected support for the Cygwin x86-32 port, which got lost at release 3.0.
--- a/backend/ValueDomain.v
+++ b/backend/ValueDomain.v
@@ -2634,11 +2634,9 @@ Qed.
 Definition cmpu_bool (c: comparison) (v w: aval) : abool :=
   match v, w with
   | I i1, I i2 => Just (Int.cmpu c i1 i2)
-  | Ptr _, I _ => cmp_different_blocks c
-  | I _, Ptr _ => cmp_different_blocks c
+  | Ptr _, I i => if Int.eq i Int.zero then cmp_different_blocks c else Btop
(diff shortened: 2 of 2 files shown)


fajb:
I think that #208 could be reproduced by  `(uintptr_t) &global - 0x1234 == 0`
To prevent any such behaviour, the validity of the pointer should be verified.
This could actually be feasible for Gl and Stk -- at some implementation cost.

I noticed this trying to prove that the value analysis should only return Bnone or Btop when cmpu is None (providing some typing constraints and ruling out Vundef). 

While doing the proof, I also (re)-discovered that:
- the equality of pointers is only defined if the pointers are valid even when the blocks are the same. Even if it is compliant with the C standard, it seems unnecessary.
- the comparison with 0 e.g. 0 < ptr(b,o) could also be defined providing validity checks but here I guess it deviates even further from the standard.

xavierleroy:
```
    (uintptr_t) &global - 0x1234 == 0`
````
I will draw the line here.  For the CompCert back-end the expression above is indistinguishable from
```
    (char *) &global - 0x1234 == NULL`
````
and you and I and everyone agrees that a pointer derived from a global can point anywhere within this global but cannot be NULL.  Otherwise you can just kiss goodbye to any hope of doing any kind of pointer analysis.

The one thing I could possibly be convinced to do is to actually distinguish the two expressions above in the back-end, e.g. by materializing conversions between integers and pointers as an operation whose dynamic semantics is the identity, but whose static semantics for value analysis is to return "top". 

> the equality of pointers is only defined if the pointers are valid even when the blocks are the same. Even if it is compliant with the C standard, it seems unnecessary.

Trust me, it is necessary.  Without the validity condition, two out-of-bounds pointers to two different blocks compare not equal, while after injection of the two blocks into one, they compare equal.  

> the comparison with 0 e.g. 0 < ptr(b,o) could also be defined 

Probably but I'd rather not encourage C programmers to write non-standard-compliant comparisons between pointers; it seems they already have a tendency to do so.

xavierleroy:
Positive feedback received, I'll merge and @fajb and I will discuss his proof attempts privately.

PR #210, opened 2017-12-07
Fix check-proof target of the Makefile after merge of Coq #6277.

ppedrot:
We simply fully qualify the modules. This is backward compatible.


maximedenes:
Would it be possible to merge this? It makes our CI builds fail, after the Coq PR was integrated a bit too quickly due to a technical glitch. Thanks!

xavierleroy:
Thanks for the fix.  Merged!

PR #217, opened 2018-01-16
Don't depend on the judgmental behavior of Bool.eqb in flocq

JasonGross:
This change is backwards compatible, and makes flocq compatible with
https://github.com/coq/coq/pull/6596.

Corresponding commit in my fork of flocq is https://github.com/JasonGross/flocq/commit/db356aa5ea0fd0234e3872f70e8972086055cd57 .

### DIFF
--- a/flocq/Appli/Fappli_rnd_odd.v
+++ b/flocq/Appli/Fappli_rnd_odd.v
@@ -199,7 +199,7 @@ case (Zeven (Zfloor r));  simpl; ring.
 apply trans_eq with (Zeven (Zceil r)).
 rewrite Zceil_floor_neq.
 rewrite Zeven_plus.
-simpl; reflexivity.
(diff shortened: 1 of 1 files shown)


xavierleroy:
Thanks for the patch, Jason.  Globally I'm in favor (of not depending too much on the actual implementation of `Bool.eqb`).  But I'll wait until @silene and the other Flocq authors apply the patch to the master Flocq sources.  The copy of Flocq that is part of CompCert might go away eventually, so it is important that it is in sync with the master sources.

JasonGross:
@xavierleroy The patch has been applied to flocq master: https://gitlab.inria.fr/flocq/flocq/commit/db356aa5ea0fd0234e3872f70e8972086055cd57

xavierleroy:
Very good, I'm merging, then.

JasonGross:
Thanks!

PR #224, opened 2018-03-09
Do not use “Require” inside sections

vbgl:
Who knows what a “Require” within a section means…

This will be deprecated in the next version of Coq.


xavierleroy:
Thanks for the heads-up.  I merged the PR manually so as to fix RTLgenproof.v in a slightly different way.

PR #228, opened 2018-04-16
update Makefile so it also installs VO files

ildyria:
Proposed to solve : https://github.com/AbsInt/CompCert/issues/227


ildyria:
I'm bumping this PR as I think this would be a big improvement for the VST developers and users. We would directly be able to install CompCert via opam to access the required `.vo` files.

xavierleroy:
See #232 for a more complete implementation of the same ideas.  Comments welcome there.

ildyria:
I like #232 , it is more flexible. :+1:

PR #229, opened 2018-04-20
Unify definitions of agree_callee_save

jeremie-koenig:
Allocproof and Stackingproof introduce incompatible definitions of `agree_callee_save`, but it turns out they can be unified with very little effort.

For what it's worth this shows up in my ongoing compositional compilation work where I use `agree_callee_save` to characterize the calling conventions used at module boundary.

### DIFF
--- a/backend/Allocproof.v
+++ b/backend/Allocproof.v
@@ -1317,15 +1317,6 @@ Proof.
   eauto.
 Qed.
 
-Definition callee_save_loc (l: loc) :=
-  match l with
-  | R r => is_callee_save r = true
-  | S sl ofs ty => sl <> Outgoing
-  end.
-
-Definition agree_callee_save (ls1 ls2: locset) : Prop :=
-  forall l, callee_save_loc l -> ls1 l = ls2 l.
-
 Lemma return_regs_agree_callee_save:
   forall caller callee,
   agree_callee_save caller (return_regs caller callee).
@@ -1345,7 +1336,7 @@ Proof.
   unfold no_caller_saves, callee_save_loc; intros.
   exploit EqSet.for_all_2; eauto.
   hnf. intros. simpl in H1. rewrite H1. auto.
-  lazy beta. destruct (eloc q). auto. destruct sl; congruence.
+  lazy beta. destruct (eloc q); auto.
 Qed.
 
 Lemma val_hiword_longofwords:
--- a/backend/Conventions.v
+++ b/backend/Conventions.v
@@ -103,3 +103,16 @@ Proof.
   generalize (loc_arguments_bounded _ _ _ H0).
   generalize (typesize_pos ty). omega.
 Qed.
+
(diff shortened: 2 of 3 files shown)


xavierleroy:
I'm still looking at this issue and I'm still unhappy about it!  

The Allocproof definition of agree_callee_save is the correct one, I think, because we really must make provisions for the caller to overwrite Outgoing stack slots.  This can be experssed in the LTL / Linear semantics by setting those locations to Vundef after a call.  And the proofs in Allocproof still go through.  It's the proofs in Stackingproof that break, and I still haven't figured out how to adapt them, and it makes me cranky.

How are you using agree_callee_save in your work?  Do you really need Outgoing slots to be preserved?

jeremie-koenig:
The short answer is: I don't need the Outgoing slots to be preserved, just for `agree_callee_save` to be the same for Allocation and Stacking. As of now, even that may no longer be required, although it would certainly be nice in principle, and reassuring.

To give more context, I define a semantics for open modules, similar to what happens in Compositional CompCert. One twist is that each language can have its own interface in terms of how function calls and returns are described, both incoming (for functions that the module implements) and outgoing (for the external calls that the module performs). Then each pass has its own "calling convention", which relates the call/return pairs of the source language to that of the target language.

Without getting into the details, I need `agree_callee_save` when I define the calling conventions for Allocation and Stacking, to express the requirement that external calls preserve the callee-save registers (and conversely the guarantee that the compiled module does too). At some point I needed to use the same definition in these two places; some changes I made over the past month make me think this should no longer be necessary.

By the way, a related issue is that the semantics of external calls in low-level languages leave the registers mostly unchanged (`rs' = set_pair (loc_result (ef_sig ef)) res rs`). It would be more realistic to set the caller-save locations to `Vundef` (in Mach and below that could include some in-memory locations if Outgoing arguments count as caller-save). I think that could count as a bug in the specification of CompCert—maybe I should create a separate report.

xavierleroy:
I agree that external calls should be formalized as destroying all caller-save registers.   That should have been done a long time ago.  I'll push a fix shortly.  Thanks for having pointed it out.

For LTL it is easy to say that Outgoing slots are also destroyed, but no so for Mach and Asm, because those slots are just memory areas in the stack frame...  This brings us back to the previous discussion, namely that Outgoing slots are morally destroyed across calls, but enforcing this policy is hard!

xavierleroy:
Concerning external calls and caller-save registers, here is the new formalization: https://github.com/AbsInt/CompCert/commit/c58b421571b0354eea602adbdae674bd1f4847e3 .  Let me know if something looks wrong.

jeremie-koenig:
This looks fine to me.

xavierleroy:
I made progress on Outgoing slots being caller-save, see #237.  Comments most welcome!

xavierleroy:
I think the need is addressed by commit 7a5aef7 .

Issue #227, opened 2018-04-16
Install does not install .vo files


ildyria:
.vo files are used with VST.

https://github.com/coq/opam-coq-archive/pull/292 changed the path for the libs but the `makefile` does not take care of the installation of the `.vo`. It would be a nice addition to help the development with VST.

lastland:
Hi, what do CompCert people think about this PR? If accepted, it would be very useful for people who want to install VST via opam.

m-schmidt:
Thanks for the PR, but we are currently busy with other things. Please be patient a little longer.

PR #232, opened 2018-05-26
Install Coq development (.vo files) if requested

xavierleroy:
This PR addresses #227 and is an alternative to #228.

`.vo` files are installed if configure options `-install-coqdev` or `-clightgen` or `-coqdevdir` are given.

Installation directory is `$(PREFIX)/lib/compcert/coq` by default and can be changed by configure option `-coqdevdir`.

All `.vo` files are installed, not just those useful for `clightgen`-generated Coq files, because other projects may use other parts of CompCert.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ else
 ARCHDIRS=$(ARCH)_$(BITSIZE) $(ARCH)
 endif
 
-DIRS=lib common $(ARCHDIRS) backend cfrontend driver debug\
+DIRS=lib common $(ARCHDIRS) backend cfrontend driver \
   flocq/Core flocq/Prop flocq/Calc flocq/Appli exportclight \
   cparser cparser/validator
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,7 +21,7 @@ else
 ARCHDIRS=$(ARCH)_$(BITSIZE) $(ARCH)
 endif
 
-DIRS=lib common $(ARCHDIRS) backend cfrontend driver debug\
+DIRS=lib common $(ARCHDIRS) backend cfrontend driver \
   flocq/Core flocq/Prop flocq/Calc flocq/Appli exportclight \
```
xavierleroy: To explain this change: there are no .v and .vo files in directory debug/, so having it in DIRS is not necessary, and causes an installation error because `debug/*.vo` doesn't exist.

xavierleroy:
@bschommer or @m-schmidt : could one of you please okay this PR?

Note that the .vo installation uses POSIX shell scripting and will probably not work with a pure Mingw/cmd.exe system.  Let me know if that's a problem in practice.

bschommer:
Works for our internal build and can be merged.

PR #235, opened 2018-06-01
solve Menhir compatibility with 20180530

ildyria:
> New command line option `--coq-lib-no-path` to suppress the above behavior
and retain the previous (now-deprecated) behavior, that is, produce
unqualified references the modules in Menhir's support library.

Thus this pull request adds a version check in `configure` to enable the option `--coq-lib-no-path` if `menhir` is of version greater or equal to 20180530.

As this is a depreciated behavior, I am aware this is more of a fix and this should be properly updated in the future. But as is, Compcert will simply not build with the current version of menhir provided by opam.

Another solution would be to limit the version of menhir at 20180528 maximum.


xavierleroy:
@jhjourdan would you please handle this issue, in the short term as well as in the longer term?

xavierleroy:
Thanks for the PR.  I adapted it in order to add backward compatibility on top of #236.

PR #237, opened 2018-06-05
Treat Outgoing stack slots as caller-save in LTL and Linear semantics

xavierleroy:
Some function arguments are passed on stack instead of in registers.  In the caller function, these arguments are put in stack locations of the "Outgoing" kind.  In the callee, they appear as stack locations of the "Incoming" kind.

Code generated by CompCert never stores to "Incoming" stack locations.  Hence the "Outgoing" stack locations of the caller look like they are preserved across the call.  However this may not be true if the called function was compiled by another compiler, and the ABI allows the callee to overwrite the parameter stack locations.  This is the case for the x86-32 ABI, and GCC-generated code often does this.

This PR revises the semantics of the LTL and Linear intermediate languages so that Outgoing stack slots are set to Vundef on return from a function call, modeling the fact that the callee could write into those stack slots.

Adapt Stackingproof to this new semantics.  This requires tighter reasoning on how Linear's locsets are related at call points and at return points.

Most of this reasoning was moved from Stackingproof to Lineartyping, because it can be expressed purely in terms of the Linear semantics, and tracked through the `wt_state` predicate.

As proposed in #229, factor out and into Conventions.v: the notion of callee-save locations, the `agree_callee_save`predicate, and useful lemmas on `Locmap.setpair`.  Now the same `agree_callee_save` predicate is used in Allocproof and in Stackingproof.


xavierleroy:
@jeremie-koenig @mansky1: if you see something problematic in this change, now is the time to report it.  I'll merge very soon otherwise.

mansky1:
Thank you for bringing it to my attention! I believe this change is orthogonal to the entry-point definitions I've been working on, so it's fine by me.

xavierleroy:
OK, thanks for the feedback.  I'll merge now.

PR #239, opened 2018-06-20
Typo: -iquore should be -iquote

fajb:


xavierleroy:
Well spotted, thanks.

PR #242, opened 2018-07-10
allow coq.8.8.1

ildyria:
Well, it compiles :)


xavierleroy:
It seems to pass the tests as well :-)  

Manually merged in https://github.com/AbsInt/CompCert/commit/679ecfeaa24c0615fa1999e9582bf2af6a9f35e7 so as to update the configure error messages as well.

PR #246, opened 2018-08-24
Import prim token notations before using them

JasonGross:
This is required for compatibility with
https://github.com/coq/coq/pull/8064, where prim token notations no longer
follow `Require`, but instead follow `Import`.

c.f. https://github.com/coq/coq/pull/8064#issuecomment-415493362

Most changes were made via
https://gist.github.com/JasonGross/5d4558edf8f5c2c548a3d96c17820169#file-fix-py
The changes to `backend/SplitLong.vp` were made manually, and the scope
opening in `cfrontend/SimplExpr.v` and `cfrontend/SimplLocals.v` were
also done manually.

### DIFF
--- a/backend/Allocation.v
+++ b/backend/Allocation.v
@@ -12,6 +12,7 @@
 
 (** Register allocation by external oracle and a posteriori validation. *)
 
+Require Import Coq.Strings.String.
 Require Import FSets FSetAVLplus.
 Require Import Coqlib Ordered Maps Errors Integers Floats.
 Require Import AST Lattice Kildall Memdata.
--- a/backend/CSE.v
+++ b/backend/CSE.v
@@ -13,6 +13,7 @@
 (** Common subexpression elimination over RTL.  This optimization
   proceeds by value numbering over extended basic blocks. *)
 
+Require Import Coq.Strings.String.
 Require Import Coqlib Maps Errors Integers Floats Lattice Kildall.
 Require Import AST Linking.
 Require Import Values Memory.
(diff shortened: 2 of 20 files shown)


xavierleroy:
Thanks for the early warning and for the proposed changes.  I'll wait for confirmation that https://github.com/coq/coq/pull/8064 is merged.  

On CompCert's side, would it work if we just put `Require Export String` in Coqlib.v, given that most if not all CompCert modules already `Require Import Coqlib`?

JasonGross:
> On CompCert's side, would it work if we just put `Require Export String` in Coqlib.v, given that most if not all CompCert modules already `Require Import Coqlib`?

Yes, that would work just as well.  Do you want me to update this PR to do that, or will you do that independently?

xavierleroy:
Good to know, thanks.  I'm not sure yet which style I prefer, but at any rate I'll take care of it, no need for you to do more.

JasonGross:
Great, thanks!  By the way, this should happen before https://github.com/coq/coq/pull/8064 is merged, as the change is backwards compatible, and all indications point to https://github.com/coq/coq/pull/8064 being merged in 8.9.  (And let me know when you do make the change, so I can remove the overlay from https://github.com/coq/coq/pull/8064.)

JasonGross:
Please merge this (or the alternative #250) ASAP.  I believe https://github.com/coq/coq/pull/8064 is blocked primarily on this and VST merging the overlays, and the branching for v8.9 is happening soon.  I expect https://github.com/coq/coq/pull/8064 to be merged within a week or so, assuming this and the VST overlay get merged.

PR #250, opened 2018-08-27
Import prim token notations before using them (Alternative to #246)

JasonGross:
This is required for compatibility with
https://github.com/coq/coq/pull/8064, where prim token notations no longer
follow `Require`, but instead follow `Import`.

c.f. https://github.com/coq/coq/pull/8064#issuecomment-415493362

Alternative to https://github.com/AbsInt/CompCert/pull/246 as per https://github.com/AbsInt/CompCert/pull/246#issuecomment-415695502

Closes #246

### DIFF
--- a/cfrontend/SimplExpr.v
+++ b/cfrontend/SimplExpr.v
@@ -26,6 +26,7 @@ Require Import Csyntax.
 Require Import Clight.
 
 Local Open Scope string_scope.
+Local Open Scope list_scope.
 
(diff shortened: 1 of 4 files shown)


xavierleroy:
Thanks for the alternate implementation, which is just fine.

Manually merged as commit f6f537d .  The other commit (whitespace removal) was unrelated, so I prefer to leave it out.  If we want to remove trailing whitespace we'll do it globally.

JasonGross:
Thanks!

Issue #244, opened 2018-08-19
configure should remove Makefile.config if given invalid arguments


andrew-appel:
Please have the configure script remove Makefile.config before building a new one.  I just wasted an hour because (1) I already had a Makefile.config, then (2) I did ` ./configure x86_64-cygwin` , which is not currently a permitted option.  So yes, ./configure produced an error message, but I missed it because "make" seemed to work just fine.

xavierleroy:
It makes a lot of sense.  See pull request #251 for an implementation + some other improvements to configure's error handling.

PR #251, opened 2018-08-28
Improve behavior of configure script in case of configuration errors

xavierleroy:
* As suggested in #244: remove leftover Makefile.config before configuration, so that if an error occurs during configuration, there is no Makefile.config file and `make` will fail.

* Configuration errors are hard to read because the full "usage" text is so long that the actual error message scrolls off the screen.  Instead, print a  short (3-line) error message, with a suggestion to do `./configure -help`.

* Print a clearer error message for unknown options (arguments starting with `-`).

### DIFF
--- a/configure
+++ b/configure
@@ -86,6 +86,10 @@ Options:
   -ignore-coq-version  Accept to use experimental or unsupported versions of Coq
 '
 
+#
+# Remove Leftover Makefile.config (if any)  (GPR#244)
+#
+rm -f Makefile.config
 
 #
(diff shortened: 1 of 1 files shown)


xavierleroy:
@m-schmidt or @bschommer : would you like to have a look at this PR?

bschommer:
Looks good to us.

xavierleroy:
Thanks.  It's a small change at any rate, so I'll merge it now.

PR #257, opened 2018-09-12
flocq: minor cleaning

vbgl:
This avoids relying on generated names (which may change in a future version of `zify`).

### DIFF
--- a/flocq/Appli/Fappli_IEEE.v
+++ b/flocq/Appli/Fappli_IEEE.v
@@ -597,9 +597,8 @@ revert H1. clear -H2.
 rewrite Zpos_digits2_pos.
 unfold fexp, FLT_exp.
 generalize (Zdigits radix2 (Zpos mx)).
-intros ; zify ; subst.
-clear -H H2. clearbody emin.
-omega.
+clearbody emin.
+intros ; zify ; omega.
 Qed.
 
 Theorem abs_B2R_lt_emax :


xavierleroy:
Dear Vincent, 

If at all possible we keep CompCert's copy of Flocq consistent with the upstream master at https://gitlab.inria.fr/flocq/flocq .  Hence, changes should first be performed upstream before we merge them in CompCert.  Would you please submit this PR upstream at https://gitlab.inria.fr/flocq/flocq/issues ?

(Adding @silene for his information.)

vbgl:
Thanks for your reply, Xavier. Unfortunately, I quite don’t understand it. The “CompCert’s copy of Flocq” seems a bit (6 months) behind “upstream master”. The master branch of Flocq would not need this patch (see: https://gitlab.inria.fr/flocq/flocq/blob/master/src/IEEE754/Binary.v#L625).

Nonetheless, I understand that you do not want to maintain your own copy of Flocq. Are there plans about updating it?

xavierleroy:
Yes, there are plans about updating to the latest released version of Flocq.  Since this is a major release that changes more things than usual, it's going to take a bit of work, so I don't have an ETA.

maximedenes:
Hello @xavierleroy. In fact, we are working on a few improvements on `zify`, like https://github.com/coq/coq/pull/8462. Unfortunately, this one (and others) break this fragile bit of Flocq.

It would be nice if we could still include these improvements on the Coq side, but we don't want to break CompCert's master. Do you have any suggestion on how we could proceed?

Maybe Vincent's patch is small enough that it could be integrated in CompCert's copy of Flocq as an exception. It should not modify any observable behavior, apart from making the script more robust, so should not get in the way of a future merge of Flocq's master in CompCert.

xavierleroy:
Thanks @maximedenes for the explanations, very useful.

In general I can't change CompCert preventively for every experimental development on Coq, especially those that have no clear release plans.

I'll make an exception for this PR because it can be viewed as anticipating a future upgrade to Flocq 3.0.0.

vbgl:
Thank you very much.

Zimmi48:
>In general I can't change CompCert preventively for every experimental development on Coq, especially those that have no clear release plans.

We try to keep the master branch of Coq quite usable so usually when a PR is merged or about to be merged in master, it means it will end up as part of a future release (unless some major regressions are found and the PR is reverted, but this does not happen frequently).

We try to ensure a good compatibility story (or at least clearly highlighting breaking changes) by testing every new change with external projects such as CompCert. This also means that we always guarantee that your project will be compatible with new Coq releases from day one. Being in Coq CI is supposed to be a deal between the Coq development team and the project's maintainers (see https://github.com/coq/coq/tree/master/dev/ci#information-for-external-library--plugin-authors). However, nobody ever asked you what you thought about this because CompCert was added in the first round of CI, before this policy was even written. So now it becomes time to ask: do you wish CompCert to be in Coq CI? If you are tired to receive these PRs from the development team, it would be entirely fine to remove CompCert from our testing suite. In fact, it wouldn't mean losing much on our side because we also test the copy of CompCert that is part of VST and I can't remember ever breaking CompCert without breaking VST. Furthermore, VST maintainers have no trouble having their copy of CompCert diverging from upstream.

maximedenes:
> VST maintainers have no trouble having their copy of CompCert diverging from upstream

Really? Where did you get that from?

Zimmi48:
From experience: they used to depend on an older copy of CompCert that they would maintain to keep compatibility with more recent versions of Coq for instance. Nowadays they are following upstream more tightly I think but still merge our PRs without questions.

Zimmi48:
OK apparently this is no longer true: https://github.com/PrincetonUniversity/VST/pull/286#issuecomment-420694919

Zimmi48:
I talked to JHJ today who told me about the particular need for stability of the CompCert code base. I can understand this concern and how our current workflow is detrimental to it. **It is very important that we ensure that this does not become a burden** for project maintainers like @xavierleroy. In particular, even if we consider that testing CompCert is very important to guarantee that we don't break things accidentally, and without wanting to come back to the old way where a completely out-of-sync copy was maintained by the development team, we could also (for instance) have an intermediary solution which would be to track a different branch (e.g. coq-ci) that tracks Coq master and to which we submit our patches and that @xavierleroy merges to master only when he wants to. Note that there are already quite a few projects in Coq CI that work like this. As for VST, if they are now keeping up-to-date with CompCert, it could be time to implement coq/coq#6639.

xavierleroy:
Thanks @Zimmi48 for the analysis, which I mostly agree with.  

As I said above I have no time and no desire to change the CompCert code base just to support experimental developments of Coq -- even on a specific CompCert branch.  The earlier I can integrate a change is when the corresponding breaking feature is 1- merged in Coq's master, and 2- scheduled for the next release (if only I could figure this out).  I can also elect to wait for the beta release of Coq.  I made a couple of exceptions lately but these will remain exceptions.

You mention going back to the old way.  The very old way I remember is that shortly before a Coq release a core Coq developer (generally Pierre Letouzey) would port CompCert's trunk to the new version and send me the diffs.  That was really nice.

Then came the not-so-old way where as you say an old version of CompCert was maintained in some Coq repo somewhere.  That was rather useless indeed.

There might be a reasonable compromise between those two old ways.

Zimmi48:
Recall that there were a few releases of Coq that were quite hard on users because they broke compatibility in many ways (the last one was 8.5 and even then I was still too fresh in Coq to really assess this by myself, so I'm mostly repeating what I've been told).
Then the development team tried hard to assess compatibility breaking changes better and earlier in the process and that was the point of having continuous integration that tests a bunch of external projects for each pull request. We also tried to make the Coq master branch mostly stable and directly usable for those users who want to. This was quite successful I think.

That being said, we can totally try to find a middle-ground were *some* projects are tested continuously and some others that require more stability are tested only when the next release gets closer. Then it would be the role of the RM to assess the breaking changes on a larger set of projects (this is already what @silene has been doing for 8.9 BTW) and decide which change go in and which ones are reverted in the stable branch.
Another solution would be to have the RM play a more active role in the merging process by having them give a formal approval each time a PR requires fixing external projects beyond plugins (which we are still forced to break all the time).
I'll try to raise these questions at the next working group.

In any case, I record that you prefer not to be in the Coq CI if it means having to merge frequent PRs from the development team. For the moment, we'll keep CompCert because it's useful to have until the next time we break it, but next time someone submits a PR, please ping me and it might be the time we end up removing CompCert from the CI.

ejgallego:
Let me briefly weight in here as likely the main person at fault for Coq's annoying CI policy, and the proximity of the WG.

First, I am not sure I understand Xavier words in the way Théo does. His complain has indeed quite a bit merit because we are asking for a change that _we don't even know if it will happen_ I didn't infer from Xavier's words any complaint about merging changes that have landed on master.

IMVHO the policy w.r.t. to Coq's CI should be:

- To aim for no breaking change without a deprecation phase. If it works in 8.X without warnings it should work in 8.X+1
- Developments willing to remain in the CI should support compilation with Coq's `master`, which means a certain level of maintenance.

One of the reasons the system is IMHO working reasonably well is that it is simple. Let's not complicate it more. The current system has nice properties wrt latency and consistency so lets enjoy that.

A very special case is ML plugins, for better or worse, the ML API is unstable and there won't be compatibility guarantees in a while. If you wanna keep a plugin updated, it must be done _à la_ Linux.

PR #259, opened 2018-10-10
Add modulus_gt_one

jbaum98:
Although the `Integers` modules include the proof `modulus_pos: modulus > 0`, I found useful the slightly stronger property that `modulus > 1`. Because it's pretty easy to prove and is useful (at least to me), it seems like we should include the proof.


xavierleroy:
Apologies for leaving this suggestion hanging.  I agree it's a good property to have.  I merged manually in commit 8c173538 so as to use the new property `modulus_gt_one` to simplify the proof of the old property `modulus_pos`.

xavierleroy:
Manually merged, closing.

PR #261, opened 2018-12-23
Do not rely on `Refine Instance Mode`

maximedenes:
This option is soon going to be turned off by default. See https://github.com/coq/coq/pull/9270

For reference, the motivation for the Coq change is to make it possible to know statically which commands open proofs and which do not.

This patch should be backward compatible, but this remains to be tested (I tested it only against Coq's `master`).

### DIFF
--- a/common/Linking.v
+++ b/common/Linking.v
@@ -59,18 +59,22 @@ Inductive linkorder_fundef {F: Type}: fundef F -> fundef F -> Prop :=
   | linkorder_fundef_refl: forall fd, linkorder_fundef fd fd
   | linkorder_fundef_ext_int: forall f id sg, linkorder_fundef (External (EF_external id sg)) (Internal f).
 
-Instance Linker_fundef (F: Type): Linker (fundef F) := {
+Program Instance Linker_fundef (F: Type): Linker (fundef F) := {
   link := link_fundef;
   linkorder := linkorder_fundef
 }.
-Proof.
-- intros; constructor.
-- intros. inv H; inv H0; constructor.
-- intros x y z EQ. destruct x, y; simpl in EQ.
+Next Obligation.
+  constructor.
+Defined.
+Next Obligation.
+  inv H; inv H0; constructor.
+Defined.
+Next Obligation.
+  destruct x, y; simpl in H.
 + discriminate.
-+ destruct e; inv EQ. split; constructor.
-+ destruct e; inv EQ. split; constructor.
-+ destruct (external_function_eq e e0); inv EQ. split; constructor.
++ destruct e; inv H. split; constructor.
++ destruct e; inv H. split; constructor.
++ destruct (external_function_eq e e0); inv H. split; constructor.
 Defined.
 
 Global Opaque Linker_fundef.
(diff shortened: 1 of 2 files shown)


xavierleroy:
Hi Maxime,

Thanks for the early warning and the proposed fix, but I'm not happy with the proposed fix...

My view is that nobody should be forced to look at proof scripts -- which are folded by default in the HTML rendition of the CompCert sources, by the way.  There should be enough explicit definitions outside "Proof...Defined" to let developers and reviewers understand what's going on.  

The "refine instance mode" you want to get rid of is perfect in this respect: I can write explicit definitions for the fields that matter and leave the rest to the proof script.  

The `refine` tactic is very bad in this respect: it hides important definitions in proof scripts that nobody should be forced to read.

Could I use `Program Instance` or some other variant of `Instance` that tells the system there might be proofs ahead?

maximedenes:
Hi Xavier,

Thanks for the feedback! Indeed I hesitated when doing the fix, so I understand your view.

I see two ways to improve the patch: one is to use `Program Instance`, which is fine for the purpose of this change, since it never opens a proof (`Next Obligation` always does). The second would be to define a new `Refine Instance` command, which would both take a term and always open a proof. When the proof term is already complete, it would just take a `Qed`/`Define` to conclude it.

The only drawback I can see to the second approach is that the command won't be available in older versions of Coq. So the first option is easier in terms of backward compatibility (from CompCert's point of view), so I'll give it a try first.

maximedenes:
@xavierleroy here's an updated attempt, using `Program Instance`. Due to `Program`'s automation, I had to tweak a bit the scripts. But the spec / proof separation is the same as before. Let me know if it looks acceptable. Thanks!

common/Linking.v
```diff
@@ -207,18 +214,24 @@ Inductive linkorder_def {F V: Type} {LF: Linker F} {LV: Linker V}: globdef F V -
       linkorder v1 v2 ->
       linkorder_def (Gvar v1) (Gvar v2).
 
-Instance Linker_def (F V: Type) {LF: Linker F} {LV: Linker V}: Linker (globdef F V) := {
+Set Refine Instance Mode.
+
```
xavierleroy: Do we really need this `Set Refine Instance Mode` now that `Program Instance` is used?

maximedenes: > it made me miss a few other occurrences

In fact, only one. I fixed it, removed `Set Refine Instance Mode` and removed some `intros` steps that have become unnecessary (the `Program` machinery performs them automatically).

So the patch should ready, but needs to be tested against the Coq version CompCert officially supports, before being merged (not expecting any issue, but you know the animal can be surprising).

ejgallego:
@maximedenes we should keep in mind the problems `Program Instance` has tho; in particular with regards to async / terminators. IMHO the potential for regressions with this migration path is not zero.

maximedenes:
@ejgallego You're right, I thought about it. I did not systematically use `Program Instance` to port code. For developments like CompCert who need some backward compatibility, I think the best migration path is first to `Program Instance`, then to a `Refine Instance` command which would be closer to `Instance`, but always opens a proof.

If we introduce `Refine Instance` in 8.10, I expect CompCert to be able to use it one or two versions later (at least the time for it to stabilize). But I'm not fully aware of the policy CompCert follows w.r.t. Coq versions compatibility these days.

PR #262, opened 2019-01-09
Compatibility with primitive integers

vbgl:
Dear CompCert developers,

Coq will soon have primitive integers (i.e., 63-bits machine arithmetic as part of the logic); see https://github.com/coq/coq/pull/6914 for details. This subsumes and replaces the 31-bit integers facility.

CompCert — and more specifically its parser — is one of the rare users of these legacy 31-bit integers. The Coq backend of menhir has been updated (thanks to Jacques-Henri) to generate code that is compatible with both old and new versions of Coq.

Coq with primitive integers provides a few compatibility modules to emulate the old 31-bit integers. Unfortunately, the extraction of old and new 31-bit integers is slightly different. Therefore, to make CompCert compatible with present and future versions of Coq, this PR does the following changes :
    - the extraction declarations that are specific to the 31-bit machine integers are moved to a dedicated module `ExtractInt31`;
    - there are two implementations of this module: one as before, an other for emulated integers using 63-bit primitive integers;
    - at configure-time, one of these modules is selected, based on the version of Coq;
    - the OCaml library includes a module `Uint63`, derived from the one of Coq, that implements the primitive operations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,11 @@ else
 ARCHDIRS=$(ARCH)_$(BITSIZE) $(ARCH)
 endif
 
+ifeq ($(COQ_HAS_PRIMITIVE_INTEGERS),true)
+EXTRACTION31DIRS=extraction_31_63
+else
 EXTRACTION31DIRS=extraction_31_31
+endif
 
 DIRS=lib common $(ARCHDIRS) backend cfrontend driver \
   flocq/Core flocq/Prop flocq/Calc flocq/Appli exportclight \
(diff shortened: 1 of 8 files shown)


maximedenes:
@xavierleroy Does this look good to you?

xavierleroy:
PS.  I could also use some explanations and background on the choice and implementation of the `uint63` type.  The reason Menhir for Coq uses the `int31` type is that it maps to an OCaml integer type that is fast and unboxed.  `uint63`, as implemented in the attached OCaml code, is neither fast nor unboxed.  Why should we use it?

PR #272, opened 2019-02-11
Make the checker happy

vbgl:
Small tweaking of the proof scripts so that the produced proof terms can be checked by the checker.

This is not to claim that there is no issue on the Coq side. But being able to run `coqchk` on *all* of CompCert seems rather nice.

Tested with Coq versions 8.8.2, 8.9.0, master (https://github.com/coq/coq/commit/3352a5b7c4507ff8fda1f5aeba83f2e141cb7a3e).

Closes https://github.com/coq/coq/issues/3026

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,12 +279,8 @@ distclean:
 check-admitted: $(FILES)
 	@grep -w 'admit\|Admitted\|ADMITTED' $^ || echo "Nothing admitted."
 
-# Problems with coqchk (coq 8.6):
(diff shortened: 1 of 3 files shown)


xavierleroy:
Very nice, thanks for the detective work!  Let me check what the best time is to merge this improvement.

xavierleroy:
As it turns out, the best time to merge is now!  So, it's merged.  Thanks again.

PR #273, opened 2019-02-13
Upgrade embedded version of Flocq to 3.1.

silene:
Main changes to CompCert outside of Flocq are as follows:

- Flocq is now compiled on its own. This will make it easier to switch to
  an external Flocq later, if wanted. This also avoids any potential name
  conflict with CompCert, since Flocq is now imported using option -Q.

- Minimal supported version of Coq is now 8.7, due to Flocq requirements.

- Most modifications are due to Z2R being dropped in favor of IZR and to
  the way Flocq now handles NaNs.

- CompCert now correctly handles NaNs for the Risc-V architecture
  (hopefully).


xavierleroy:
Thanks a lot for this upgrade!  

We can't merge immediately because of some special CompCert release in preparation but will do just after, in a few weeks.

Dropping support for Coq 8.6.1 is no problem at all, I was considering doing it anyway.

Concerning the way Flocq is built: from your description I was expecting `$(MAKE) -C flocq`, i.e. just use Flocq's own build procedure.  From a quick look at the diffs it looks more complicated than that.  Did I miss something?

silene:
From a CompCert user/developer point of view, there is no change. You just type `make whatever` as before. The main difference is that Flocq and CompCert (minus Flocq) are now built using separate logical paths. The critical lines are the following ones. (Other changes to `Makefile` are just to make the experience painless for the user.)

    COQINCLUDES=$(foreach d, $(RECDIRS), -R $(d) compcert.$(d)) -Q flocq compcert.flocq
    $(FLOCQ:.v=.vo): COQINCLUDES=-R flocq compcert.flocq

So, the day you want to make CompCert depend on an already installed Flocq, you just have to replace the `From compcert.flocq` directives in CompCert's Coq files by `From Flocq` and to remove the then useless lines from `Makefile`. I could have moved all the relevant rules to a separate `Makefile`, but I see little gain in that.

xavierleroy:
> I could have moved all the relevant rules to a separate `Makefile`, but I see little gain in that.

I was hoping Flocq's original (or lightly-modified) Makefile could have been used for that purpose, so that CompCert's toplevel Makefile knows almost nothing about Flocq beyond the need to do `$(MAKE) -C flocq`.  

I know recursive makefiles should be used sparingly, but the single-Makefile approach here is quite heavy, with two `.depend` files and what not.

silene:
The thing is, Flocq does not use any Makefile.

xavierleroy:
> Flocq does not use any Makefile.

Well, it has its own build routine.  But I noticed it spends a lot of time (5 minutes or so) compiling Pff.v, a module we don't need in CompCert.  So perhaps a build customized for CompCert is better, in the end.

xavierleroy:
I finally had time to review carefully.

First of all: thanks a lot for having done this porting work, I would have had a hard time doing it myself!

Second: I now understand better the various possibilities for the build procedure.  One thing I realized is that the old Makefile (with -R for all directories, including flocq/) still works, provided Flocq modules are imported without qualifiers (`Require Import Core`) rather than with a qualifier (`From compcert.flocq Require Import Core`).  We all agree that `-R` + unqualified imports is more prone to conflicts and less self-documented than `-Q` + qualified imports.  It's just that I would prefer to switch the whole of CompCert to the `-Q` style, at some point in the future, instead of having a transition period with a mixture of `-R` and `-Q`.

The other thing I realized is that @silene is right :-) : just sticking a `From Flocq` in 3 files is enough to use an existing installation of Flocq.  It's magical :-)  And it's good to know because some users have good reasons to want to do this.  What would be even better is to be able to select an existing Flocq at configure-time, without changing the Coq sources, just by passing different options to coqc. I haven't succeeded yet,  although something like the following comes close:
```
coqc -R Flocq `coqc -where`/user-contrib/Flocq
```
Comments and suggestions most welcome.  In the meantime I'm moving towards keeping the old Makefile and the old `-R` approach outlined above, and merging the rest of this PR.

xavierleroy:
PR #283 fleshes out my proposal above, with more conservative changes on Makefile.  Comments most welcome!

xavierleroy:
This PR evolved into #283 and was merged, so, closing.

PR #275, opened 2019-02-25
Maximum supported Menhir version.

jhjourdan:
The Coq backend of Menhir will soon enjoy a large refactoring, making it incompatible with the version of MenhirLib currently in CompCert. This PR adds a check in configure to make sure that the version of Menhir is not more modern than the current one (20181026).

### DIFF
--- a/configure
+++ b/configure
@@ -555,12 +555,13 @@ fi
 
 MENHIR_REQUIRED=20161201
 MENHIR_NEW_API=20180530
+MENHIR_MAX=20181026
 menhir_flags=''
 echo "Testing Menhir... " | tr -d '\n'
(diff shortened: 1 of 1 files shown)


xavierleroy:
Actually the latest version of Menhir seems to be 20181113.  That's what OPAM provides, and it works with CompCert trunk.  @jhjourdan is it OK to have 20181113 as the upper bound in the configure script?

jhjourdan:
Yes. I have no idea why my `opam` does not provides me this version...

xavierleroy:
Make sure you're using OPAM 2.  Recent packages are not available in OPAM 1.

PR #276, opened 2019-02-26
New parser based on new version of the Coq backend of Menhir

jhjourdan:
This is the new version of the parser, based on recent updates of Menhir.

The corresponding changes in Menhir are not yet released: to test these changes, you should install the current master branch of Menhir. This is the reason why this PR is marked WIP. But apart from that (and updating `configure`), I think this PR is ready.

Changes include:

1. A rewrite of the Coq interpreter of Menhir automaton, with dependent types removing the need for runtime checks for the well-formedness of the LR stack. This seem to cause some speedup on the parsing time (~10% for lexing + parsing).

2. Thanks to 1., I was able to no longer use int31 for comparing symbols: Since this is only used for validation, `positive`s are enough.

3. Speedup of Validation: on my machine, the time needed for compiling Parser.v goes from about 2 minutes to about 1 minute. I am not sure of the actual reason for this, but this seem to be related to a performance bug I detected in the completeness validator and to the use of `positive` instead of `int31` (I used to use a weird hack for representing `int31` constants for making sure they were evaluated efficiently after extraction... but this could actually be rather slow when executed with `vm_compute`).

4. Menhir now generates a dedicated inductive type for (semantic-value-carrying) tokens (in addition to the already existing inductive type for (non-semantic-value-carrying) terminals. The end result is that the OCaml support code for the parser no longer contain calls to `Obj.magic`. The bad side of this change is that the formal specification of the parser is perhaps harder to read.

5. The parser and its library are now free of axioms (I used to use axiom K and proof irrelevance for easing proofs involving dependent types).


jhjourdan:
Note that I also updated the `LICENSE` file, reflecting that coq-menhirlib is under LGPLv3 (like Flocq), and that I moved it to the root of the repository.

xavierleroy:
What about the use of streams?  Cf. #231 .  Will these changes to Menhir and its supporting library remove dependencies on the Stream library from Coq?

jhjourdan:
You are right, I forgot about this. I need a bit more work to get rid of the Stream dependency.

jhjourdan:
> What about the use of streams? Cf. #231 . Will these changes to Menhir and its supporting library remove dependencies on the Stream library from Coq?

The dependency to the Stream library of Coq is now removed, Menhir now uses its own type for input buffer. This type is a negative coinductive type, so that it should not be affected by coq/coq#7536.

jhjourdan:
I added another change: the fuel parameter is now specified using the logarithm (in base 2) of the maximum number of steps needed. In practice, we use the value 50, so that we are sure that we will not run out of fuel in reasonable computation time. From the perspective of CompCert, the advantage is that this prevents us from using the `let rec inf = S inf` hack.

configure
```diff
@@ -553,30 +553,8 @@ else
...
-          echo "Error: CompCert requires a version of Menhir between $MENHIR_REQUIRED and $MENHIR_MAX, included."
-          missingtools=true
-      fi;;
-  *)
-        echo "NOT FOUND"
-        echo "Error: make sure Menhir version $MENHIR_REQUIRED or later is installed."
-        missingtools=true;;
-esac
+# TODO
+menhir_includes="-I `menhir --suggest-menhirLib`"
```
xavierleroy: Remind me to fill in something there before I merge :-)

cparser/Lexer.mll
```diff
@@ -490,97 +487,95 @@ and singleline_comment = parse
...
+      | Pre_parser.SIGNED loc -> loop (Parser.SIGNED loc)
+      | Pre_parser.SIZEOF loc -> loop (Parser.SIZEOF loc)
+      | Pre_parser.SLASH loc -> loop (Parser.SLASH loc)
+      | Pre_parser.STAR loc -> loop (Parser.STAR loc)
+      | Pre_parser.STATIC loc -> loop (Parser.STATIC loc)
+      | Pre_parser.STRING_LITERAL (wide, str, loc) ->
           (* Merge consecutive string literals *)
           let rec doConcat wide str =
             try
               match Queue.peek tokens with
```
xavierleroy: On a completely unrelated topic: this code would benefit from the "match exception" syntax.
```
match Queue.peek tokens with
| ... ->
| ... ->
| exception Queue.Empty ->
```

cparser/Parser.vy
```diff
@@ -36,7 +36,7 @@ Require Import List.
   CHAR SHORT INT LONG SIGNED UNSIGNED FLOAT DOUBLE CONST VOLATILE VOID
   STRUCT UNION ENUM UNDERSCORE_BOOL PACKED ALIGNAS ATTRIBUTE ASM
 
-%token<cabsloc> CASE DEFAULT IF ELSE SWITCH WHILE DO FOR GOTO CONTINUE BREAK
+%token<cabsloc> CASE DEFAULT IFt ELSE SWITCH WHILE DO FOR GOTO CONTINUE BREAK
```
xavierleroy: I guess `IF` now conflicts with something?  Coq's notation? Something else?

Nothing serious but I would have a slight preference for `IF_` over `IFt`.

cparser/Parser.vy
```diff
@@ -114,10 +114,10 @@ primary_expression:
 | var = VAR_NAME
     { (VARIABLE (fst var), snd var) }
 | cst = CONSTANT
-    { (CONSTANT (fst cst), snd cst) }
+    { (Cabs.CONSTANT (fst cst), snd cst) }
```
xavierleroy: I'm a bit sad that some Cabs constructors  are qualified with `Cabs.` (because they must be) but others are not.  This could make this grammar slightly harder to read than if all constructors were consistently prefixed.

cparser/Parser.vy
```diff
@@ -573,61 +573,52 @@ gcc_attribute_word:
...
 
 direct_declarator:
 | id = VAR_NAME
     { Name (fst id) JUSTBASE [] (snd id) }
 | LPAREN decl = declarator RPAREN
     { decl }
 | decl = direct_declarator LBRACK quallst = type_qualifier_list expr = assignment_expression RBRACK
-    { match decl with Name name typ attr loc =>
-        Name name (ARRAY typ (rev' quallst) (Some (fst expr))) attr loc end }
+    { let 'Name name typ attr loc := decl in Name name (ARRAY typ (rev' quallst) (Some (fst expr))) attr loc }
```
xavierleroy: Long line...  Split after the `in` ? 
(6 other occurrences below)

jhjourdan:
The corresponding version of Menhir (20190613) have been released, I rebased/squashed the commits, and tested the whole thing.

As far as I can tell, this is now ready to merge.

xavierleroy:
Thank you.  I'm preparing the CI machines here at Inria so that they can test this new version.

xavierleroy:
With Coq 8.7.2 under Cygwin, I get
```
16:54:35 File "./MenhirLib/Validator_complete.v", line 232, characters 6-62:
16:54:35 Error:
16:54:35 Anomaly
16:54:35 "File "plugins/ltac/tacinterp.ml", line 1157, characters 35-41: Assertion failed."
```
Will try other Coq versions.

jhjourdan:
Thanks for the test. Let me look into this more in detail.

xavierleroy:
After upgrading to Coq 8.9.1, still under Cygwin, I now get
```
COQC cparser/Parser.v
File "./cparser/Parser.v", line 217, characters 20-2939:
Error:
In environment
x : terminal
The term
 "match x with
  | ADD_ASSIGN't => 1
  | ALIGNAS't => 2
[...]
end" has type "nat" while it is expected to have type
"positive".
```
Coq 8.8.2 under Linux works fine, though.  

I'll investigate on my side too.

jhjourdan:
The compatibility issue with 8.7.2 is now fixed in this branch (and in upstream Menhir).

The compatibility issue with 8.9.1 required a change in Menhir itself (in the generator, not in the support library). The fix will be shipped with the next Menhir release:

https://gitlab.inria.fr/fpottier/menhir/commit/5f195649fda98fda58c83e4ce956b41a0db4fec6

I hope @fpottier will not mind too much doing yet another Menhir release :D Thanks, @fpottier!

jhjourdan:
And sorry for still being a young Padawan too naive about compatibility between Cod versions! I should have checked that.

fpottier:
I'll be happy to make a new release (now that the release scripts work again, it is reasonably easy).

@jhjourdan, could you modify the opam descriptions of the two versions that we have released already (20190613 and 20190620) so as to document their compatibility constraints? This will avoid problems if someone installs one of these versions. Thanks!

jhjourdan:
This PR should now be fixed for both Coq 8.7 and Coq 8.9.

xavierleroy:
The tests are OK, and we agreed to merge, so let's merge!  

Thanks for all the hard work.

PR #277, opened 2019-03-09
update the comment of the free operation

chaomaer:
I think there is a mistake in the comment. It should be `freeable` not `writable`


xavierleroy:
Well spotted, thanks.  I'll merge as part of the next round of changes on the master sources.

PR #278, opened 2019-03-11
Simplify ec_mem_extends

jeremie-koenig:
Preservation of "out of bounds" locations is not required.

### DIFF
--- a/arm/Asmgenproof.v
+++ b/arm/Asmgenproof.v
@@ -742,7 +742,7 @@ Opaque loadind.
   generalize (transf_function_no_overflow _ _ H3); intro NOOV.
   exploit builtin_args_match; eauto. intros [vargs' [P Q]].
   exploit external_call_mem_extends; eauto.
-  intros [vres' [m2' [A [B [C D]]]]].
+  intros [vres' [m2' [A [B C]]]].
   left. econstructor; split. apply plus_one.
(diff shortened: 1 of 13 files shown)


xavierleroy:
There was a time where the preservation of "out of bounds" locations was important: when the Stacking pass (if I remember correctly) could be formalized with a memory extension.  Later it was turned into a memory injection, so as to support relocation of stack offsets, so the properties of injections were used instead.   

But the point remains that this preservation of "out of bounds" locations is part of the interface for memory extensions, in my opinion.  Do you have specific reasons to get rid of this property?  Other problems with memory extensions, more generally?

jeremie-koenig:
No, I was just going over the precise requirements of each pass and noticed it was unused. I don't have a reason to get rid of it beyond the slight decrease in complexity.

One thing I noticed that you may want to change is that RTLgenproof.v:707,739 destruct the property one level too deep (which doesn't end up mattering in practice); more generally it may also be useful to replace the unused component with `_` in destruct patterns so that it's immediately apparent that the `Mem.unchanged_on` part isn't used by the proof. (I can do this if you have any interest)

xavierleroy:
> One thing I noticed that you may want to change is that RTLgenproof.v:707,739 destruct the property one level too deep

Well spotted, thanks a lot!  This is fixed by commit f7a4a677 .

PR #283, opened 2019-03-26
Upgrade embedded version of Flocq to 3.1, continued

xavierleroy:
This is a follow-up to pull request #273.  The only difference is that there are fewer changes to Makefile, because we keep the old approach of using `-R` for the `flocq` directory, instead of using `-Q`.


xavierleroy:
For their information: @silene @bschommer @m-schmidt .

silene:
I am fine with `-R`. My reason for `-Q` was to detect potential conflicts. Indeed, module names from Flocq 3 have become quite bland, e.g., `Binary`, so I wanted to make extra sure I was not introducing any conflict.

By the way, changes to the `.gitignore` file are no longer needed.

xavierleroy:
Thanks for the quick reply.  I agree with the blandness of the module names :-) and the need for `-Q` in the future.  Will fix the .gitignore too.

bschommer:
The changes work fine in our build and testing infrastructure, thanks for the work on this.

Just a minor note, it seems to be the case that ````.nra.cache```` is missing in the ````.gitignore````.

xavierleroy:
Thanks @bschommer for the re-test and the hint about `.gitignore`.  There are a lot of cache files that Coq can generate :-).  Merging now...

Issue #282, opened 2019-03-23
Provide version information in in preprocessor


nemequ:
I'd like to add support for CompCert to [Hedley](https://nemequ.github.io/hedley/), but AFAICT there is no way to tell from the preprocessor which *version* of CompCert is in use, so there is no way to tell when certain features can be enabled.

For example, it looks like noinline was added with version 3.2, so if I'd like to be able to do something like

```c
#if defined(__COMPCERT__) && (__COMPCERT__ >= 30200)
#  define HEDLEY_NO_INLINE __attribute((noinline))
#else
#  define HEDLEY_NO_INLINE
#endif
```

It looks like this would basically just require something like

```diff
diff --git a/driver/Frontend.ml b/driver/Frontend.ml
index 88b47854..30d78592 100644
--- a/driver/Frontend.ml
+++ b/driver/Frontend.ml
@@ -19,7 +19,7 @@ open Driveraux
 
 let predefined_macros =
   [
-    "-D__COMPCERT__";
+    "-D__COMPCERT__=30500";
     "-U__STDC_IEC_559_COMPLEX__";
     "-D__STDC_NO_ATOMICS__";
     "-D__STDC_NO_COMPLEX__";
```

But obviously the version number shouldn't be hard-coded, and I'm not sure where to get it.

xavierleroy:
I agree this would be nice.  Thanks for the suggestion, I'll look into it.

xavierleroy:
A proposed implementation: #284.

PR #284, opened 2019-03-26
Define macros with CompCert's version number

xavierleroy:
As suggested in #282, it can be useful to `#ifdef` code depending on specific versions of CompCert.

Assuming a version number of the form `MM.mm` ~~or `MM.mm.pp`~~, the following macros are predefined:

`__COMPCERT_MAJOR__=MM`       (the major version number)
`__COMPCERT_MINOR__=mm`       (the minor version number)
~~`__COMPCERT_PATCHLEVEL__=pp`  (the patchlevel, or 0 if none)~~
`__COMPCERT_VERSION__=MMmm` (two decimal digits each, e.g. 305 for version 3.5)

Having 3 macros for MAJOR, MINOR and PATCHLEVEL is what GCC and Clang do.  (Edit: but during the discussion we found PATCHLEVEL useless.)

Having the combined VERSION macro is convenient to write tests such as
```
#if __COMPCERT_VERSION__ >= 302
// use feature introduced in CompCert 3.2
#endif
```

### DIFF
--- a/driver/Frontend.ml
+++ b/driver/Frontend.ml
@@ -11,6 +11,7 @@
 (*                                                                     *)
 (* *********************************************************************)
 
+open Printf
 open Clflags
 open Commandline
 open Driveraux
@@ -35,7 +36,7 @@ let v_number =
 (* Predefined macros: version numbers, C11 features *)
 
 let predefined_macros =
-  Printf.([
+  let macros = [  
     "-D__COMPCERT__";
     sprintf "-D__COMPCERT_MAJOR__=%d" v_major;    
     sprintf "-D__COMPCERT_MINOR__=%d" v_minor;    
@@ -46,7 +47,10 @@ let predefined_macros =
     "-D__STDC_NO_COMPLEX__";
     "-D__STDC_NO_THREADS__";
     "-D__STDC_NO_VLA__"
-  ])
+  ] in
+  if Version.buildnr = ""
+  then macros
+  else sprintf "-D__COMPCERT_BUILDNR__=%s" Version.buildnr :: macros
 
 (* From C to preprocessed C *)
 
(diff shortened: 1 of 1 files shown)


m-schmidt:
There is a slight collision with our version numbers which look like `18.10` or `18.10i` (2 digits for year and month plus optional `i` for intermediate releases). With our version number scheme `__COMPCERT_PATCHLEVEL__` will always be zero.

Therefore it would be nice to add an additional define, e.g. `__COMPCERT_BUILDNR__`, which maps to `Version.buildnr`. Our build number is currently in the 4 million range but the integer range should suffice in the foreseeable future.

m-schmidt:
Thanks, the extension looks good.

xavierleroy:
You're right about `__COMPCERT_BUILDNR__`, I added a definition if the build number exists.  

Concerning the `18.10` version numbers, I was not aware of this problem.  Do you think the presence of `__COMPCERT_BUILDNR__` is enough to distinguish between the two series of version numbers?

Finally, I'm not sure the PATCHLEVEL part is useful.  For the public version the only cases where we've used 3-level version numbers was to adapt to several versions of Coq, e.g. 3.0 was for Coq 8.5 and 3.0.1 was for Coq 8.6 but otherwise functionally identical, so there was no point in announcing this difference to the C code that is being compiled.  

Getting rid of the patchlevel part also makes the `__COMPCERT_VERSION__` macro easier to read, e.g. `305` instead of `30500`.

Opinions welcome!

m-schmidt:
I think the presence of `__COMPCERT_BUILDNR__` is sufficient to distinguish the two version number schemes, should if it ever be necessary. We also don't see a great benefit for an explicit patch level. It's easier to simply increment the minor version.

xavierleroy:
OK, we're very much in agreement.  I removed the PATCHLEVEL and simplified VERSION accordingly.  The code also gets simpler.

xavierleroy:
I have a feeling that we are all happy about this PR, so let me merge ASAP.

PR #286, opened 2019-04-02
Use Z.log2 instead of the deprecated log_inf

vbgl:
Function `log_inf` from the `Zlogarithm` module has been deprecated for a few years. It will be removed from the standard library (see https://github.com/coq/coq/pull/9811).


xavierleroy:
Thanks for the early warning.  I pushed a slightly different fix (with no extra lemma in Coqlib.v) that should cut the dependency to Zlogarithm just as well.

This said, I will complain about the lack of deprecation warnings on Zlogarithm.

PR #292, opened 2019-05-16
Fix a typo in a comment in Csyntax.v

btj:


xavierleroy:
Thanks for noticing and for the patch.

PR #293, opened 2019-05-17
Branchless implementations of unsigned int32 <-> float conversions for x86

xavierleroy:
Based on a suggestion by Rémi Hutin.
The new implementation produces smaller code (3 FP instructions and 2 integer "and") and has no conditional branch, which is good for security and for WCET.
On a microbenchmark, no significant change in execution speed was observed.


bschommer:
I did run our tests successfully with this branch.

xavierleroy:
I have doubts about my micro-benchmark.  I'll try to benchmark some more and report here.

bschommer:
Thanks for doing the benchmark work.

xavierleroy:
I did more measurements on several x86 processors of various age.  Globally, the new, branchless implementation is slower than the old implementation, even on pseudorandom inputs where I was expecting the conditional branch to be poorly predicted!  The mysteries of modern processors...

My conclusion is that this branchless implementation cannot become the default.  However, it could be selected by a command-line flag such as `-ffavor-branchless` as proposed in #297.

I'll return to this PR when #297 is merged.

xavierleroy:
As promised, the new, branchless implementation of `floatofintu` is now used if the flag `-ffavor-branchless` is given; by default, the old, branch-full implementation is used.

Following the same approach, I added the branchess implementation of `intuoffloat` (the reverse conversion) that was proposed by @RemiHutin in #295.  Again, because it is not a complete performance win, the old implementation remains the default, and the branchless implementation is used if `-ffavor-branchless` is given.

atrieu:
On a related note, the handwritten assembly for converting float to unsigned long (and others in runtime/x86_64, and probably others) are not branchless if I understand ASM correctly.
Is there anything similar to this PR planned about that ?

xavierleroy:
You or anyone who's good with x86 asm can try to produce efficient branchless implementations of the functions in runtime/x86_64/*.S.  But it's not going to be pretty.

xavierleroy:
See #306 for a much simpler approach, although it is branchless on x86-64 bits only.

PR #297, opened 2019-06-01
If-conversion optimization

xavierleroy:
Support for "conditional move" instructions was recently added to the CompCert back-end.

This pull request adds an "if-conversion" optimization that converts some conditional statements (`if...else...`) and conditional expressions (`... ? ... : ...`) into these conditional move instructions, thus generating branchless code for those conditionals.

The expressions and statements that are converted are of the form
```
            cond ? a1 : a2    
            if (cond) { x = a1; } else { x = a2; }
            if (cond) { x = a1; }
            if (cond) { /*skip*/; } else { x = a2; }
```
with the following constraints:
* `a1` and `a2` are "safe" expressions, containing no operations that can fail at run-time, such as memory loads or integer divisions
* `x` is a local variable whose address is not taken
* the target architecture supports "conditional move" at the type of `a1` and `a2`.

A heuristic deactivates if-conversion when it looks potentially costly, e.g. if `a1` is much more costly to evaluate than `a2`, or conversely.  The command-line option `-ffavor-branchless` overrides this heuristic and generates conditional moves whenever semantically valid.  The command-line options `-fno-if-conversion` and `-O0` turn if-conversion off entirely.


xavierleroy:
Should be of interest to @bschommer @m-schmidt @monniaux @RemiHutin .

xavierleroy:
A note on the `-f[no-]if-conversion` and `-ffavor-branchless` options:

An alternative I considered is to have a single `-fif-conversion` option with three values: off, normal, and aggressive.  But it doesn't fit the pattern of other optimizations (`-fopt` / `-fno-opt`).  Plus, the `-ffavor-branchless` option could be used elsewhere to select instruction forms that are branch-less but possibly more costly than the current branch-full forms, as in #293 and #295.

test/regression/ifconv.c
```diff
@@ -0,0 +1,129 @@
+#include <stdio.h>
+
+/* Several equivalent forms that should be turned into cmov */
```
curiousleo: I was wondering if you could clarify this (for my benefit, this is not actually a PR review comment): from reading this test, my understanding is that it checks whether code involving the relevant pattern (`b ? x : y` and `if b then x else y`) has the correct observable behaviour. As far as I can tell, this test says nothing about when the optimisation implemented in this PR is (or ought to be) triggered - that is, this test should succeed even on a version of CompCert without the optimisation. Did I get that right?

xavierleroy: Yes, you're entirely right.  To see where the optimization triggered, I look at the generated assembly code.

backend/Cminortyping.v
```diff
@@ -0,0 +1,798 @@
...
+      type_stmt tret e s1
+  | Sexit n =>
+      OK e
+  | Sswitch sz a tbl dfl =>
+      type_expr e a (if sz then Tlong else Tint)
+  | Sreturn opta =>
+      match tret, opta with
+      | None, None => OK e
+      | Some t, Some a => type_expr e a t
+      | _, _ => Error (msg "inconsistent return")
```
bschommer: This actually triggers a warning in the frontend, maybe we should turn the warning in an error, since in the frontend we still have line information.

bschommer:
I did run our tests on this and it works fine.

monniaux:
Ported to mppa-k1c.

backend/Cminortyping.v
xavierleroy: Well spotted!  You're right, code such as
```
double f(void) { return; }
```
triggers this error in Cminortyping, when it is too late.

I'm pretty sure the typing in Cminortyping can be relaxed to accept this case, while preserving the desired type soundness property.  Let me check ASAP.

xavierleroy: Commit d5e7d2b relaxes the typing rule for "return" without argument.

xavierleroy:
Squashed, rebased, re-tested.  Ready for merging.

PR #301, opened 2019-07-04
Avoid relying on `Export` bug

maximedenes:
It seems that CompCert was relying on `Export` sometimes having no effect, see https://github.com/coq/coq/issues/10480

I'm preparing a fix for this Coq bug. This patch makes CompCert compatible with the fix, and is backward compatible (tested on Coq's `master`).

### DIFF
--- a/lib/Floats.v
+++ b/lib/Floats.v
@@ -24,6 +24,7 @@ Require Import Program.
 Require Archi.
 
 Close Scope R_scope.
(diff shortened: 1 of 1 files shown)


xavierleroy:
> It seems that CompCert was relying on Export sometimes having no effect

Purely by accident!  

Thanks for the proposed patch, it looks OK to me.

Issue #300, opened 2019-07-03
OCaml 4.08.0


ejgallego:
Dear CompCert devs,

we have started testing CompCert with OCaml 4.08.0 in Coq's CI, however a few warnings due to the use of the `Pervasives` module have popped up, and make the build fail.

What would be your preferred approach to handle these?
- disable deprecation warnings in the corresponding places
- unqualify the use of `Pervasives` functions

TIA, Emilio

bschommer:
A PR removing the warnings would be appreciated, since I cannot test the new OCaml version at the moment since there is still no camlp5 release compatible with OCaml 4.08.

ejgallego:
> A PR removing the warnings would be appreciated, since I cannot test the new OCaml version at the moment since there is still no camlp5 release compatible with OCaml 4.08.

Ok, we'll give it a go. camlp5 is compatible with 4.08.0 since a few days, I guess I'll push a new package to OPAM as to help some Coq plugins that still use it.

bschommer:
Yes, but it is also still missing a release, I know that the current master should work with camlp5 but a release is needed for our internal build structure.

xavierleroy:
Thanks for noticing us, Emilio.

I had a look at the uses of `Pervasives`:
* One occurrence of `Pervasives.snd` in a context where there is no risk of confusion with the `snd` extracted from Coq.  So, unqualified `snd` will work just fine.  Besides, the `snd` extracted from Coq would work too :-)
* Seven occurrences of `Pervasives.compare`.  Again, removing the qualification works fine.  It just makes the code a little harder to read in one occurrence.

So, I'm OK with removing the qualification.  When 4.08 is the older version we support, we'll have the option to reintroduce `Stdlib.compare` in some places.

I'll submit a PR shortly.

xavierleroy:
Pull request: #302 .

ejgallego:
Thanks a lot @xavierleroy !

For info, I think that Coq will support OCaml 4.08.0 officially in the 8.11 version [expected at the end of the year], but indeed we want to start testing now so your patch helps a lot.

xavierleroy:
Patch was merged in master sources.

ejgallego:
Thank you @xavierleroy !

PR #302, opened 2019-07-05
Do not use `Pervasives.xxx` qualified names

xavierleroy:
This is a fix for issue #300.  Starting with OCaml 4.08, `Pervasives` is deprecated in favor of `Stdlib`, and uses of `Pervasives` cause fatal warnings. This commit uses unqualified names instead, as no ambiguity occurs.


xavierleroy:
I could test with OCaml 4.08.0 and Coq 8.9.1 compiled with an earlier version of OCaml.  There were other warnings-as-errors, concerning `open` statements and shadowing of module names.  I added a second commit to this PR to fix this.  Now CompCert compiles fine with OCaml 4.08.0.

xavierleroy:
Approved privately.  Merging.

PR #303, opened 2019-07-05
Fix compatibility with Coq 8.10 and later.

jhjourdan:
Menhirlib was reported by the opam repository maintainers as being incompatible with Coq master (see https://github.com/coq/opam-coq-archive/pull/821).

This PR fixes that incompatibility, hopefully not introducing any incompatibility with former Coq versions.

### DIFF
--- a/MenhirLib/Validator_classes.v
+++ b/MenhirLib/Validator_classes.v
@@ -17,7 +17,7 @@ Require Import Alphabet.
 
 Class IsValidator (P : Prop) (b : bool) :=
   is_validator : b = true -> P.
-Hint Mode IsValidator + -.
+Hint Mode IsValidator + - : typeclass_instances.
 
 Instance is_validator_true : IsValidator True true.
 Proof. done. Qed.
@@ -31,7 +31,7 @@ Proof. done. Qed.
 
 Instance is_validator_and P1 b1 P2 b2 `{IsValidator P1 b1} `{IsValidator P2 b2}:
   IsValidator (P1 /\ P2) (if b1 then b2 else false).
-Proof. split; destruct b1, b2; auto using is_validator. Qed.
+Proof. by split; destruct b1, b2; apply is_validator. Qed.
 
 Instance is_validator_comparable_leibniz_eq A (C:Comparable A) (x y : A) :
   ComparableLeibnizEq C ->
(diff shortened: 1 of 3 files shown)


SkySkimmer:
I guess this is due to https://github.com/coq/coq/pull/9160 or at least something close to it.

jhjourdan:
FWIW, these changes have been integrated upstream. They are not yet released upstream, but we do not really care, since CompCert bundles them.

MenhirLib/Validator_safe.v
```diff
@@ -184,8 +184,8 @@ Instance impl_is_state_valid_after_pop_is_validator state sl pl P b :
   IsValidator (state_valid_after_pop state sl pl -> P)
               (if is_state_valid_after_pop state sl pl then b else true).
 Proof.
-  destruct (is_state_valid_after_pop state0 sl pl) eqn:EQ.
-  - intros ??. auto using is_validator.
+  destruct (is_state_valid_after_pop _ sl pl) eqn:EQ.
```
ejgallego: Here you may want to keep the name, but explicitly use `state0` in the binder.

ejgallego:
Interesting, at least for Coq master only the `state0` problem seems to be present.

MenhirLib/Validator_classes.v
```diff
@@ -31,7 +31,7 @@ Proof. done. Qed.
 
 Instance is_validator_and P1 b1 P2 b2 `{IsValidator P1 b1} `{IsValidator P2 b2}:
   IsValidator (P1 /\ P2) (if b1 then b2 else false).
-Proof. split; destruct b1, b2; auto using is_validator. Qed.
+Proof. split; destruct b1, b2; by apply is_validator. Qed.
```
ejgallego: A nit on ssreflect's style, `by` is supposed to appear first in the chain of tactics:
```coq
by split; destruct b1 b2; apply is_validator.
``

jhjourdan:
> Interesting, at least for Coq master only the state0 problem seems to be present.

This is the only problem which makes the compilation fail. The other problems only trigger warnings.

MenhirLib/Validator_safe.v
jhjourdan: I don't really care about the proof script, but I think `state` is a better name in the statement of the theorem.

MenhirLib/Validator_classes.v
jhjourdan: Thanks.

jhjourdan:
> The other changes make less sense to me. What is wrong with auto using... that Coq 8.10 feels obliged to warn about?

The warning was about the `Hint Mode` which was not explicitly specifying a hint database. And indeed, I made a mistake there, since I wanted to change the `typeclass_instances` database instead of the `core` one.

With this fixed, the TC resolution mechanism becomes (willingly) weaker, and some of the `eauto`-based proof scripts were failing.

xavierleroy:
Thanks for the explanation, it makes perfect sense now.  Merging!

PR #305, opened 2019-07-07
Fix #304 Changes in configure broke Windows build

### LINKED ISSUE #304: Changes in configure broke Windows build
Dear CompCert Team,

a recent change in the configure file, I think it was 1df887f5a275e4c31096018ff1a8fdfc39bca591, broke the Windows build. The menhir library path ended up to contain a carriage return. I will provide a PR to fix this. This also hardens configure for menhir installations which return a path with backward slashes.

Best regards,

Michael

MSoegtropIMC:
This PR fixes issue #304, that is the build of CompCert on Windows (for inclusion of CompCert as optional addon in the Coq Windows Installer).

### DIFF
--- a/configure
+++ b/configure
@@ -560,7 +560,7 @@ case "$menhir_ver" in
   20[0-9][0-9][0-9][0-9][0-9][0-9])
       if test "$menhir_ver" -ge $MENHIR_REQUIRED; then
           echo "version $menhir_ver -- good!"
-          menhir_include_dir=`menhir --suggest-menhirLib`
+          menhir_include_dir=$(menhir --suggest-menhirLib | tr -d '\r' | tr '\\' '/')
           if test -z "$menhir_include_dir"; then
               echo "Error: cannot determine the location of the Menhir API library."
               echo "This can be due to an incorrect Menhir package."


xavierleroy:
Thank you for the fix.

MSoegtropIMC:
@xavierleroy : would you mind if I choose the hash just after this PR as version to be included in Coq 8.10? I prefer to use a tag with at most some minor local build patches, but CompCert 3.5 is not Coq 8.10 compatible, so I would also have to include the 8.10 compatibility patch. If it is not just Windows build compatibility and the latest tag does not work with the targeted Coq version I usually prefer a hash rather than a patched up tag. But I can't tell what is preferable for CompCert - there seem to be quite a bit of non trivial changes on top of 3.5, so it might be preferable to use a patched up version of 3.5.

xavierleroy:
@MSoegtropIMC : sorry for the late reply, I was away and could not test as I wanted.

> would you mind if I choose the hash just after this PR as version to be included in Coq 8.10?

I mind.  Please do not release working versions of CompCert, especially when there have been significant changes from the latest "official" release of CompCert (here, 3.5).

The next release of CompCert should occur at the beginning of September, and will be compatible with Coq 8.10.

I you really can't wait, you can try to take CompCert 3.5 and apply (as patches) commits 0f919eb2 and e9c13650.  It seems to make CompCert 3.5 compatible with Coq 8.10 beta 2.

MSoegtropIMC:
> I mind.

That is what I guessed. I would prefer to have a constant set of Coq developments in the Coq installer and 8.10 will be released soon, so I will follow your recommendation and apply the two patches on top of 3.5. We can include the next release of CompCert in 8.10.1. then.

Btw.: in the future we plan to have separate releases for Coq and an extended Coq platform. The Coq release will only contain Coq and a minimal set of libraries and the Coq platform release will contain an extended set of tools and should come about 6..8 weeks after the Coq release. This should make this much easier, e.g. your releases beginning of September would be perfectly in time then. So I hope that we can avoid such issues in the future.

MSoegtropIMC:
@Xavier Coq 8.10 will included CompCert 3.5 with the two cherry-picks you suggested. See (https://github.com/coq/coq/pull/10621). Thanks for the detailed description!

PR #307, opened 2019-07-11
Revised specification of NaN payload behavior

xavierleroy:
When an FP arithmetic instruction produces a NaN result, the payload of this NaN depends on the architecture.

Before, the payload behavior was specified by 3 architecture-dependent parameters: `Archi.choose_binop_pl_64` and `Archi.choose_binop_pl_32` and `Archi.fpu_results_default_qNaN`.  This was adequate for two-argument operations, but doesn't extend to FMA.

In preparation for FMA support, this commit generalizes the `Archi.choose` functions from two arguments to any number of arguments.  In passing, `Archi.fpu_results_default_qNaN` is no longer needed.

### DIFF
--- a/arm/Archi.v
+++ b/arm/Archi.v
@@ -16,7 +16,7 @@
 
 (** Architecture-dependent parameters for ARM *)
 
-Require Import ZArith.
+Require Import ZArith List.
 (*From Flocq*)
 Require Import Binary Bits.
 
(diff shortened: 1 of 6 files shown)


xavierleroy:
@silene @jhjourdan you're very welcome to have a look at this PR if you have time.

silene:
So, from reverse-engineering the proofs, it seems like:

- all the architectures have the same default NaN (positive, most significant payload bit set), except for x86 (negative);
- all the supported architectures quiesce signaling NaNs in the exact same way (by setting the most significant payload bit);
- all the architectures expand and contract NaN payloads between binary32 and binary64 the same way;
- all the architectures modify a single bit for negation and absolute value (so, no quiescing);
- the x86-64 and PowerPC architectures select the leftmost NaN (and its sign) if it exists, or return a default one if not;
- the ARM architecture selects the leftmost signaling NaN if it exists, or behaves like previous architectures otherwise;
- the RiscV architecture always generates a default NaN (compliance with IEEE 754 is overrated).

That is correct, as far as I remember.

Note that `fma(0,inf,qnan)` might well return the default quiet NaN (due to the invalid exception) rather than the third operand, which does not match the above description. That is an underspecified part of the IEEE-754 standard, unfortunately, so I am not quite sure what architectures do.

xavierleroy:
Thank you very much for the review and analysis.  What you wrote is 1- what was intended, and 2- consistent with my tests on x86, ARM and PowerPC, although those tests are never conclusive (we're observing one processor, but other processors with the same instruction set could behave differently).  As concerns FMA, I started to test and will keep the "0, inf, qnan" case in mind.

xavierleroy:
Actually, one point of your summary has an exception:

> all the architectures expand and contract NaN payloads between binary32 and binary64 the same way;

PowerPC keeps binary32 numbers expanded to binary64 in registers, so binary32 -> binary64 is a no-op and signaling NaNs rare not quiesced.  The current code accounts for this oddity.

xavierleroy:
See #309 for the addition of FMA.  The case "0, inf, qnan" that @silene warned about is indeed treated differently in ARM and in x86.

PR #317, opened 2019-10-04
Concurrency with CompCert

scuellar:
Full documentation can be found in Chapter 4 of my thesis, here: [Concurrent Permission Machine for modular proofs of optimizing compilers with shared-memory concurrency, Chapter 4](https://github.com/AbsInt/CompCert/files/3688514/thesis_ch4.pdf)

The current state of the code includes some incomplete details as detailed below:

1. The extraction doesn't work due to an unknown error. It seems like a problem in configuration.
2. We have only propagated the proofs for X86_32 and x86_64. The latter might have some errors that will be fixed in the next couple of days. Try exploring the code in 32-bit mode for now.
3. Some admitted lemmas. From printing assumptions on the final theorem:
    - `Lineartyping.wt_pre_main_locset_all` (should be straightforward)
    - `Conventions1.loc_arguments_non_rep` (should be easy)
    - `injection_injection_composition` (should be easy)
    - `Mach.get_arguments_correct` (should be straightforward)
    - `Asm.get_arguments_correct` (should be straightforward)
    - `ValueAnalysis.sound_entry`
We expect that `sound_entry` can be proven from `entry_point`, or a small modification of it. There are more comments in ValueAnalysis.c, where the lemma is admitted.
 

Challenges to integrating to CompCert master:
1. Some passes are not proven: the current proves don't cover Dead Code Elimination, Unused Globals and the two passes above Clight. In the short term, we propose to have two configurations of CompCert. For the concurrent configuration, we skip the unproven passes and get the result we have right now.  For the standard configuration, we prove that our stronger simulation implies the old simulation, and then we can have a compiler with all phases. In the long term, we expect to be able to prove Dead Code Elimination, Unused Global and SimplLocals as described in the chapter. We haven't explored adapting SimplExpr. 
2. Support for other architectures is missing.
3. The safety preservation theorem is not rebuilt, only forward simulation is proven. We prove backward simulation and safety preservation separately for our concurrent semantics. Proving the safety preservation theorem is discussed in subsection 4.5.1 of the thesis.
4. **This pull request is compatible with a commit from May 21st** (f047fcb7852ff58c0c62f10d41f91f3f88552780). That is the last time I merged CompCert's branch. We don't foresee major difficulties bringing this code up to date.


jeehoonkang:
@scuellar Thank you for sharing your code.  Which concurrency model are you using?  Do you support x86, POWER, ARM, or RISC-V architecture concurrency semantics?

scuellar:
@jeehoonkang, thank you for you interest. 

I'll send you our latest manuscript so you can follow the current standing of our work. We prove a "reordering theorem" so we can assume sequentially consistent, cooperative semantics through compilation and then derive all the other executions at the assembly level. We currently target X86 and are looking into expanding that.

xavierleroy:
For the record, here is a technical report by @scuellar and coauthors describing the approach followed in this PR in great details: https://www.cs.princeton.edu/research/techreps/TR-014-19.

This PR was a technical preview, and is not mergeable as is, so I think it is clearer if I close it now.

PR #321, opened 2019-10-31
Do not rely on a failure of automatic proof search

vbgl:
I would like to make “intuition” a bit stronger, in a way that would break this proof script.

See https://github.com/coq/coq/pull/11018 for details.

### DIFF
--- a/lib/IntvSets.v
+++ b/lib/IntvSets.v
@@ -102,7 +102,7 @@ Proof.
   simpl. rewrite IHok. tauto.
   destruct (zlt h0 l).
   simpl. tauto.
-  rewrite IHok. intuition.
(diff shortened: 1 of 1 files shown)


bschommer:
I tried it with the oldest and newest support coq version and it works, also the change is not that large and I think it would be fairly safe to merge this, what do you think @xavierleroy ?

xavierleroy:
Sorry, I missed this PR initially, and thanks for the ping.

The proposed change looks innocuous and is good style anyway.  (Normally I try to never use `intuition` alone and always use `intuition auto` or `intuition congruence` or something.)

I'll approve and merge ASAP.

PR #324, opened 2019-12-12
CompCert with comments

bcip:
We developed a mod of CompCert that preserves comments in C programs during
compilation. Our main purpose is to allow writing comments in C code as hints for
verification, e.g.
```
x = 0;
y = 1;
//@ assert x+y == 1
```
So we only allow comments to appear in position of a statement but not inside a statement or an expression. This mod may be also useful for tracking generated code by CertiCoq.

CompCert's original compilation processes, including ccomp and clightgen, consist of three phases:
1. preprocess C code using GCC;
2. parse C code to AST;
3. translate C code step by step while preserving semantics.

Our mod preserves comments until the intermediate representation Csharpminor during translation. We explain the implementation in the following sections.

For VST, preserving comments down to Clight would be sufficient; but a project of Philippa Gardner's may find it useful to preserve to Csharpminor.

#### Preprocessor
The original preprocessing phase removes comments in C code. Our modified clightgen provides an additional flag `-C` which means to preserve comments. If flag `-C` is set, clightgen will let GCC preserve comments when preprocessing. So the comments are passed to the next phase.

#### Parser
The parser first lexes C code into a sequence of tokens. Each comment is lexed into a token. A new kind of token COMMENT is added to the lexer. The data carried by a COMMENT token is a string, which is the comment's content. Single-line and multi-line comments are not differentiated by the lexer. Because the parser is not able to parse comments at every position, but comment should be allowed to appear anywhere, users are requested to put an '@' at the beginning of "annotation" comments, as in VeriFast and ACSL. The lexer only produces tokens for comments led by '@' and ignores other comments.

In the parser, comments are not defined as statements but block items. CompCert's CFG is defined as
```
 statement:
 | LBRACE block_item_list RBRACE
 | ...

 block_item_list:
 | block_item*

 block_item:
 | statement
 | declaration
 | pragma
+| comment
```
We add `comment` as a kind of block item. This prevents ill-formed programs, e.g.
```
if(b)
  /*@ comment */
  x = 1;
```

#### Intermediate representations and semantics
In each intermediate representation down to Csharpminor, we add a constructor `Scomment : comment -> statement` in the definition of statement. The semantics of comment is defined by the following rule:
```
  | step_comment_skip: forall f cmt k e le m,
    step (State f (Scomment cmt) k e le m)
      E0 (State f Sskip k e le m)
```
So it takes a step to reduce a comment to a skip. In each pass, `Scomment` is translated to `Scomment` in the next representation. We prove this translation sound for each pass.

The intermediate languages after Cminor, inclusively, do not support comments. So when translating Csharpminor to Cminor, i.e. in Cminorrgen, `Scomment` is translated to `Sskip`. To prove this translation's correctness, we set the measure of comments as 1%nat as follows, because Csharpminor's semantics takes one step to reduce a comment to a skip, but the generated Cminor program is directly a skip.
```
 Fixpoint seq_left_depth (s: Csharpminor.stmt) : nat :=
   match s with
   | Csharpminor.Sseq s1 s2 => S (seq_left_depth s1)
+  | Csharpminor.Scomment _ => S O
   | _ => O
   end.
```


atrieu:
Just by curiosity, how different is it from [CompCert's annotation mechanism](http://compcert.inria.fr/man/manual006.html#sec91) ?

xavierleroy:
Thanks for this contribution.  I agree it's nice to have a lightweight syntax based on special comments to add logical assertions to a C program, like Frama-C does with the ACSL specification language.  

However, I have several problems with your implementation, so I'm not going to use it.

1- There is no need to add `Scomment` as a new kind of AST node in C, CompCert C, Clight.  Rather, the comments should be recorded as calls to special builtin functions, like CompCert does for annotations (http://compcert.inria.fr/man/manual006.html#sec91) and for debugging information.  

What you need is a new kind of external functions (see module AST), `EF_comment`, carrying a string, taking no arguments, and having as semantics "return Vundef and produce no event in the trace" (see module Events).  

Calls to this external function propagate for free throughout the compilation chain.  No need to add cases to proofs. 

If you're afraid they might invalidate optimizations, you can remove calls to these builtin function at any time, for example while generating RTL code, since their semantics is that of a "no op".

2- The parsing is not quite right.  You should take inspiration from Frama-C and ACSL, https://frama-c.com/download/acsl.pdf, in particular section 1.2.  For example, consider:
```
  if (x == 1) //@ annotation
    x = 0;
```
I have the impression that this is parsed like
```
  if (x == 1) {
    //@ annotation
  }
  x = 0;
```
This is wrong: the `x = 0` should still be the "then" branch of the `if` statement.

Also, if I understand correctly, `@` comments will cause  parsing errors if they occur outside of statements, e.g. in declarations or outside of a function body.  Either they should be ignored, or `@` comments should not be recognized by default, but only if a special flag is given.  In its present state, your patch prevents the compilation of ACSL-annotated source C files!

bcip:
@xavierleroy 
1. We don't want to request users to write`__builtin_annot("...");`, instead we want to let them just write comments. But I agree that the implementation is an issue to discuss. It may be a better way to convert comments into external function calls when parsing instead of having a new kind of statements.

2. The parser we implemented treats this case as a parser error.
```
if (x == 1) //@ annotation
    x = 0;
```
If it is allowed to write
```
if (x == 1)
    //@ annotation
    x = 0;
```
then the correct interpretation of the following code is not obvious
```
if (x == 1)
    x = 0;
    //@ annotation
```
In other words if users want write annotation in our style, they must write inside braces. This does not prevent compilation of ACSL-annotated C code, because the comments are only parsed if `-C` flag is set, otherwise the comments are removed when preprocessing.

QinxiangCao:
@xavierleroy  @atrieu

For comparison between current CompCert's annotation mechanism and our annotations,

1. One benefit of current CompCert's annotation mechanism is: we can treat it as function calls and do not need to write extra code to handle these comment in every compilation passes.

2. One disadvantage of  current CompCert's annotation is the concern of possible degrade optimization.

Ideally, we do not want to write extra semantic definitions and transition code for these comments, and we do not want to write extra code to distinguish these comments with normal function calls. For now, I do not find one way to achieve both.

Also, current CompCert's annotation mechanism can directly mention identifiers (program variables, etc.).

-----------------

@bcip 

For annotations for the following situation,

```
if (x == 1)
    //@ annotation
    x = 0;
```

we throw a passing error now. But I guess we can improve it. A better parser should ignore comments at such position as if there is no "@" after "//".

bcip:
@QinxiangCao @xavierleroy 
I checked the problem about program like
```
if
    //@ atnnotation
    x = 0;
```
1. Our implementation reports a parsing error for this program.
2. ACSL allows this program.
3. VeriFast doesn't allow this program, either, reporting a message "pure statement not allowed here" (on a 2018 version).

I still suggest this program should not be allowed. And this will not prevent compilation of ACSL-annotated programs because comments are passed to the parser only if `-C` flag is set, otherwise the comments are removed in preprocessing. If `-C` is set, that means we are using our own annotation, not ACSL.

xavierleroy:
I agree this is a potentially interesting feature, but as explained above I disagree with the proposed implementation, so I'm closing this PR.

PR #328, opened 2020-01-15
faster signed division by 2

monniaux:
Because signed division in C rounds to 0, it cannot be implemented directly by an arithmetic shift right when the divisor is a power of two. Some architectures (PPC, K1C) have a special "round to 0 shift right" operator, some (Risc-V, ARM, AArch64) don't and the instruction is emulated by 3 or 4 instructions. This patch uses a simpler sequence of instructions when the divisor is 2, like gcc does.


xavierleroy:
This PR was discussed with the AbsInt team.  We concluded that it's too much code and proofs for a tiny optimization.  CompCert is getting bigger, now that it has 5 back-ends that need maintaining in parallel, so we'd rather focus future optimization work on optimization that have a bigger payoff.

PR #330, opened 2020-01-27
Makefile.menhir: specify menhirLib.cmxa instead of menhirLib.cmx.

fpottier:
This should allow CompCert to compile with Menhir 20200123 and newer.

### DIFF
--- a/Makefile.menhir
+++ b/Makefile.menhir
@@ -41,7 +41,7 @@ MENHIR_FLAGS = -v --no-stdlib -la 1
 # Using Menhir in --table mode requires MenhirLib.
 
 ifeq ($(MENHIR_TABLE),true)
-  MENHIR_LIBS = menhirLib.cmx
+  MENHIR_LIBS = menhirLib.cmxa
 else
   MENHIR_LIBS =
 endif


bschommer:
Indeed that should work, however this is not really backward compatible, since older menhir versions did not install the `.cmxa` file as far as I know.

fpottier:
Hi Bernhard, you are correct, older versions of Menhir installed `menhirLib.{cmi,cmo,cmx,cmxs,o}` whereas the new version installs `menhirLib.{a,cma,cmi,cmt,cmti,cmx,cmxa,cmxs}`. The old set of files was hardcoded somewhere in our Makefiles, whereas the new set of files is automatically chosen by `dune` and/or `opam`.

I suppose you could complicate `Makefile.menhir` by testing which of `menhirLib.cmx` and `menhirLib.cmxa` exists. This would allow to preserve compatibility between a new CompCert and an old Menhir.  But it may be simpler to just require a newer Menhir. (Also, the compatibility between an old CompCert and a new Menhir will remain broken, unless you apply the patch retroactively.)

xavierleroy:
> I suppose you could complicate Makefile.menhir by testing which of menhirLib.cmx and menhirLib.cmxa exists. This would allow to preserve compatibility between a new CompCert and an old Menhir.

See #331 for an implementation of this idea.  It might be worth preserving compatibility for a few versions of CompCert.

xavierleroy:
We went with the more convoluted fix at #331.  Thanks for having diagnosed the issue, anyway.

Issue #329, opened 2020-01-25
CompCert doesn't seem to compile with the last version of Menhir


giltho:
Hello !

CompCert is not compiling properly with the last version of Menhir.

Here's the configuration :
```
Testing assembler support for CFI directives... yes
    Testing Coq... version 8.10.2 -- good!
    Testing OCaml... version 4.08.1 -- good!
    Testing OCaml .opt compilers... yes
    Testing Menhir... version 20200123 -- good!
    Testing GNU make... version 3.81 (command 'make') -- good!
    
    CompCert configuration:
        Target architecture........... x86
        Hardware model................ 64
        Application binary interface.. macosx
        Endianness.................... little
        OS and development env........ macosx
        C compiler.................... gcc -arch x86_64
        C preprocessor................ gcc
        Assembler..................... gcc
        Assembler supports CFI........ true
        Assembler for runtime lib..... gcc -arch x86_64 -c
        Linker........................ gcc
        Linker needs '-no-pie'........ false
        Math library.................. 
        Build command to use.......... make
        Binaries installed in......... /usr/local/bin
        Runtime library provided...... true
        Library files installed in.... /usr/local/lib/compcert
        Standard headers provided..... true
        Standard headers installed in. /usr/local/lib/compcert/include
        Coq development will not be installed
```

With these parameters, it seems that CompCert is not building, using :
```bash
./configure x86_64-macosx #as showed above
make -j 4
```
The compilation dies in the middle. Here are the last few lines :
```
...
OCAMLOPT extraction/Memtype.ml
    OCAMLOPT extraction/PeanoNat.ml
    OCAMLOPT cparser/Machine.ml
    OCAMLC   cparser/Env.mli
    OCAMLC   cparser/Cprint.mli
    OCAMLOPT extraction/Ring.ml
    OCAMLOPT lib/Printlines.ml
    OCAMLOPT extraction/DecidableClass.ml
    OCAMLC   cparser/Transform.mli
    OCAMLOPT extraction/Cabs.ml
    OCAMLOPT cparser/pre_parser_aux.ml
    OCAMLC   cparser/pre_parser.mli
    OCAMLC   cparser/Cflow.mli
    File "cparser/pre_parser.mli", line 114, characters 10-56:
    114 |   include MenhirLib.IncrementalEngine.INCREMENTAL_ENGINE
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #331, opened 2020-01-30
Revised menhirLib autoconfiguration

xavierleroy:
This is a fix for #329, alternate to #330, in that it supports both old and new versions of Menhir.

Since Menhir version 20200123, we need to link with menhirLib.cmxa instead of menhirLib.cmx.

This commit chooses automatically the file to link with:
- menhirLib.cmxa if it exists in the menhirLib installation directory,
- menhirLib.cmx otherwise.

To reliably find the installation directory, configure was changed to record the menhirLib directory in Makefile.config, variable MENHIR_DIR, instead of a pre-cooked command-line option MENHIR_INCLUDES.

Makefile.extr was adapted accordingly.


bschommer:
Sorry for the late reply, this works for us and should fix the menhir issue.

PR #340, opened 2020-03-25
removal of frame pointer restoration for leaf functions on aarch64

monniaux:


monniaux:
Ouch. There is a problem - x30 is reused by CompCert in some cases.

xavierleroy:
As discussed with @monniaux and colleagues by e-mail: if we find it important to optimize leaf functions for multiple target architectures (beyond the simple optimization done for PowerPC already), I'd rather try to factor out some of the work in the Stacking pass so as to reduce the amount of per-target work.

The problem with x30 is noted: the semantics of built-in functions document which allocatable registers they can destroy, but not which special registers (such as x30, "return address", on ARM64, or GPR0, "temporary or 0" on PowerPC) they can destroy.  I hope to make that clearer in the Asm semantics in the near future.

PR #348, opened 2020-04-20
Installed compcert.ini also in the coq library folder

MSoegtropIMC:
Installing the compcert.ini file also in the coq library folder makes it easier and more reliable to find the file for Coq based tools like VST. The current install location in /share seems to primarily target C tools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ ifeq ($(CLIGHTGEN),true)
 	$(MAKE) clightgen
 endif
 
-proof: $(FILES:.v=.vo)
+proof: $(FILES:.v=.vo) compcert_conf.mk
 
 # Turn off some warnings for compiling Flocq
 flocq/%.vo: COQCOPTS+=-w -compatibility-notation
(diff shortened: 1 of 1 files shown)


xavierleroy:
Currently, `compcert.ini` is intended to be read by `ccomp` and no other tool, which explains its location and the fact that it doesn't contain all parameters one may wish to know about the configuration.

Probably, users of CompCert as a Coq library have different needs than `ccomp`.  Could you please tell us more about
- the information needed
- which programs are going to read that information (Makefiles, shell scripts, Coq, etc) (so that we agree on the most convenient format).

MSoegtropIMC:
I agree that it would be better to have separate files matching their specific purpose. I only came to this solution because the compcert.ini file is well suited as is, except for the missing bitsize field.

The only user of this file I am aware of the the Princeton VST makefile. It used to extract this information from CompCert's Makefile.config, but this obviously does not work in case of using system installed CompCert Coq files. The relevant lines in the VST makefile are:
```
ARCH ?= $(shell awk 'BEGIN{FS="="}$$1=="arch"{print $$2}' $(COMPCERT)/compcert.ini)
BITSIZE ?= $(shell awk 'BEGIN{FS="="}$$1=="bitsize"{print $$2}' $(COMPCERT)/compcert.ini)
```
One could read in the file directly, if it would contain `?=` assignments. If the file does not contain `?=` assignments, the formatting does not make much of a difference.

So how about having a file "compcert_conf.mk" which contains the above information and possibly a few more entries using make `?=` assignments?

MSoegtropIMC:
I pushed an alternative for discussion. This creates a file compcert_conf.mk and installs it in the coq library folder. The contents is e.g.:
```
# CompCert configuration parameters for Coq makefiles
ARCH ?= x86
BITSIZE ?= 32
MODEL ?= 32sse2
ABI ?= macosx
ENDIANNESS ?= little
SYSTEM ?= macosx
```
Would this be agreeable?

xavierleroy:
Thanks for the code. In parallel I and the AbsInt people worked on a proposal, then integrated yours.  The result is commit cc465218.  Compared with the proposal:
* The installed file is called `compcert.config` 
* The file uses `=` instead of `?=` so that it works also as a shell script and is easier to parse with other tools.
* Variable names are prefixed with  `COMPCERT_` to avoid possible name conflicts
* The file also contains variables for CompCert's version numbers. (Could be useful too, even though the version information is already installed in file VERSION.)

Let us know if something is missing.

MSoegtropIMC:
> Let us know if something is missing.

This is definitely a good compromise and should suite all needs. I will close this PR. I will merge your commit as a patch in the opam `coq-compcert.3.7~coq-platform` package.

xavierleroy:
> This is definitely a good compromise and should suite all needs

Excellent!  Thanks for the confirmation.

PR #352, opened 2020-05-01
[coq] use ListNotations explicitly

llelf:
overlay for coq/coq#11992
it's backward-compatible, can be merged now.

PR #354, opened 2020-05-03
In Coq-MenhirLib: backport upstream changes.

jhjourdan:
I.e., import ListNotations wherever it is necessary so that we do not rely on it being exported by Program.

This is the part of #352 which is specific to the MenhirLib directory. There are a few adaptations, but this should not cause incompatibilities with your PR, @llelf.

### DIFF
--- a/MenhirLib/Alphabet.v
+++ b/MenhirLib/Alphabet.v
@@ -11,7 +11,8 @@
 (*                                                                          *)
 (****************************************************************************)
 
-From Coq Require Import Omega List Syntax Relations RelationClasses.
+From Coq Require Import Omega List Relations RelationClasses.
+Import ListNotations.
 
 Local Obligation Tactic := intros.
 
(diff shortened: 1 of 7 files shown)


jhjourdan:
This is a backport from https://gitlab.inria.fr/fpottier/menhir/-/commit/53f94fa42c80ab1728383e9d2b19006180b14a78

xavierleroy:
Actually there was a problem: you're using `Declare Scope`, which was introduced in Coq 8.10, making MenhirLib incompatible with earlier Coq versions.  Since CompCert supports coq 8.8 and up, commit 2cccb812 removes the `Declare Scope` (and the makefiles already pass the right options to silence the "scope should be declared" warning).  Think about this compatibility issue for the upstream version.

jhjourdan:
Thanks for the notice !

Should I be surprised that Coq developers introduce `Declare Scope` and soon after force people to use it?

xavierleroy:
> Should I be surprised that Coq developers introduce Declare Scope and soon after force people to use it?

You should.  Not "soon after": in the same release.  Be grateful that the lack of `Declare Scope` is only a warning, which can be turned off.

PR #355, opened 2020-05-06
[Proposal] Tweak "perm_inv" in extends/inject

alxest:
The rationale behind this PR is as follows.

- We can think of a memory operation (say, free_weak) that lowers "Cur" permission without changing "Max" permission. (As described in [here](https://hal.inria.fr/hal-00703441/document))
Then, it is desirable to have lemmas like `free_weak_left_extends` and `free_weak_left_inject`, just like `free_left_extends` and `free_left_inject`.
However, such lemmas do not hold currently because of `perm_inv` conditions.
This PR tweaks the conditions so that aforementioned lemmas hold.

- Also, this form looks more like forward-simulation style and (arguably) a little simpler.

This PR compiles with Coq 8.10.1.
It turns out that no single lemma/theorem's statement is changed, except for the unused ones (`perm_extends_inv`/`perm_inject_inv`).
Thanks for reading this!

### DIFF
--- a/backend/Deadcodeproof.v
+++ b/backend/Deadcodeproof.v
@@ -41,9 +41,6 @@ Record magree (m1 m2: mem) (P: locset) : Prop := mk_magree {
   ma_perm:
     forall b ofs k p,
     Mem.perm m1 b ofs k p -> Mem.perm m2 b ofs k p;
-  ma_perm_inv:
-    forall b ofs k p,
-    Mem.perm m2 b ofs k p -> Mem.perm m1 b ofs k p \/ ~Mem.perm m1 b ofs Max Nonempty;
   ma_memval:
     forall b ofs,
     Mem.perm m1 b ofs Cur Readable ->
@@ -68,7 +65,6 @@ Lemma mextends_agree:
 Proof.
   intros. destruct H. destruct mext_inj. constructor; intros.
 - replace ofs with (ofs + 0) by omega. eapply mi_perm; eauto. auto.
-- eauto.
 - exploit mi_memval; eauto. unfold inject_id; eauto.
   rewrite Z.add_0_r. auto.
 - auto.
@@ -161,8 +157,6 @@ Proof.
   constructor; intros.
 - eapply Mem.perm_storebytes_1; eauto. eapply ma_perm; eauto.
   eapply Mem.perm_storebytes_2; eauto.
-- exploit ma_perm_inv; eauto using Mem.perm_storebytes_2.
-  intuition eauto using Mem.perm_storebytes_1, Mem.perm_storebytes_2.
 - rewrite (Mem.storebytes_mem_contents _ _ _ _ _ H0).
   rewrite (Mem.storebytes_mem_contents _ _ _ _ _ ST2).
   rewrite ! PMap.gsspec. destruct (peq b0 b).
@@ -206,8 +200,6 @@ Lemma magree_storebytes_left:
 Proof.
   intros. constructor; intros.
 - eapply ma_perm; eauto. eapply Mem.perm_storebytes_2; eauto.
-- exploit ma_perm_inv; eauto.
-  intuition eauto using Mem.perm_storebytes_1, Mem.perm_storebytes_2.
 - rewrite (Mem.storebytes_mem_contents _ _ _ _ _ H0).
   rewrite PMap.gsspec. destruct (peq b0 b).
 + subst b0. rewrite Mem.setN_outside. eapply ma_memval; eauto. eapply Mem.perm_storebytes_2; eauto.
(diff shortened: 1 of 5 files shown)


xavierleroy:
Very interesting observations, thank you!  

I can't remember why `perm_inv` was introduced: for compatibility with version 1 of the memory model?  (the one with bounds but no fine-grained  locations)? because it looked like it would make some proofs simpler?  

@lennartberinger @scuellar : do you see a problem with this proposal?

lennartberinger:
If I remember correctly, the perm-inv axioms were introduced by Santiago (but would likely also have simplified CompComp). As they express that Cur permissions can't be turned from, say, readable to writable by compilation, I suspect preserving them is crucial for Santiago's concurrency proof
(to maintain permission coherence over compilation). But he's more on top of these things these days than I am.

scuellar:
From what I see, the new `mext_perm_inv` and `mi_perm_inv` are redundant with `mi_perm` in `mem_inj`. So this PR is effectively removing `mi_perm_inv` and `mext_perm_inv`.

First, it is no surprise that you can remove those properties without breaking CompCert since CompCert's use of Cur is very limited. 

However, I consider `mext_perm_inv` and `mi_perm_inv` an important part of the interface of Cur. Those statements are saying that the compiler is not allowed to arbitrarily increase its Cur permissions. I think this is a fundamental property to satisfy external expectations over the Cur permission. In fact, it seems to me that without `perm_inv` the compiler could pump Cur = Max in the first pass and then forget about it, effectively making Cur trivial (I haven't tried this so I might be wrong. What do you think?).

In practice, my concurrency proof heavily relies on `perm_inv` to enforce that the Cur permissions of different threads still "add up" after compilation. Removing `perm_inv` will not be compatible with our approach.

Finally, it seems like, even though you can't prove `free_weak_left_extends`, you can prove a similar theorem. See how `store` deals with it in `store_within_extends` and `store_outside_extends` in Memory.v. I have done similar proofs many times.

(For completeness: the compiler is allowed to increase Cur for "new locations" such as spilled variables. This is enforced by the second disjunct ` ~perm m1 b ofs Max Nonempty` which, by requiring the Max permissions to be None ensures that the context--e.g., other threads or modules--have no assumptions about those locations. I could imagine the use of something like structured injections to express that you can increase Cur in locations that have not being shared with the context. )

xavierleroy:
Thanks a lot, @lennartberinger and @scuellar , for the context.  Indeed, "perm_inv" is a relatively recent addition that was prompted by your work (see commit fb44646eb7308c1dd6a44b85415528982083200b); I just forgot about that.

So, what does this mean for this PR?  We're not going to remove "perm_inv" entirely, as this would break the legitimate use outlined by @scuellar .  Can the use case outlined by @alxest be accommodated in another way?  

Independently, I find it interesting that "perm_inv" is not necessary for the CompCert semantic preservation proofs.  Maybe I should update the proofs as proposed in this PR, so that CompCert becomes independent (so to speak) from this condition.

alxest:
Thanks for your comments, @xavierleroy, @lennartberinger and @scuellar.
I see your points, and I agree to keep `perm_inv` as is.
(I think I can avoid my problem without changing `perm_inv`, though it required some mixture of ideas.)

Below is my understanding of the problem.

- In CompCert and CompCertM, permission violations are interpreted as undefined behavior.
So, it is natural for the target to have more permission than the source. (e.g., in physical assembly, we have full permission everywhere)

- However, in Concurrent CompCert, some permission violations are interpreted as an empty behavior (trace).
To be specific, it uses "angelic" nondeterminism, where a random oracle assigns permission to each memory location. Then, when we consider the behavior of a program, we ignore traces that are stuck because of a silly oracle. (e.g., an oracle that never gives permission to anyone) In other words, permission violation here is interpreted as empty behavior (trace).

To sum up, we have two different interpretations of permission, which effectively made it both covariant and contravariant w.r.t. translation (except the `Max None` case).
Out of curiosity, I wonder what would happen if we have two different permission data for these two different interpretations. (violating the former is undefined behavior, and the latter is empty behavior) It might satisfy both use cases.


.
.


> From what I see, the new mext_perm_inv and mi_perm_inv are redundant with mi_perm in mem_inj. So this PR is effectively removing mi_perm_inv and mext_perm_inv.

I didn't realize that, thanks!

> In fact, it seems to me that without perm_inv the compiler could pump Cur = Max in the first pass and then forget about it, effectively making Cur trivial (I haven't tried this so I might be wrong. What do you think?).

I agree that the compiler can do that, but still, Cur permission can be useful. When defining source semantics (e.g., C or mathematical spec), this can be used to give undefined behavior to ill-mannered accesses.

> Finally, it seems like, even though you can't prove free_weak_left_extends, you can prove a similar theorem. See how store deals with it in store_within_extends and store_outside_extends in Memory.v.

Unfortunately, `free_weak_left_extends`/`free_weak_left_inject` are the precise theorem I need, and I cannot add any condition to this. I have strong reasons for this but I guess it is out of topic.
I think I can avoid the problem by changing the semantics of my specification -- I used to lower/raise `Cur`permission but now I lower/raise `Max` permission. This is unnatural in CompCert but it seems to work by exploiting RUSC -- in spec verification I use a weaker version of injection (which allows `Max` perm to increase) and in compiler verification I use the normal injection.

alxest:
> Independently, I find it interesting that "perm_inv" is not necessary for the CompCert semantic preservation proofs. Maybe I should update the proofs as proposed in this PR, so that CompCert becomes independent (so to speak) from this condition.

That would definitely make my life easier. As I commented above, I guess I need to use two versions of injection (with and without `perm_inv`).

xavierleroy:
I looked at the CompCert sources again and I confirm that the `perm_inv` conditions are not used anywhere in the proofs of semantic preservation.  If you don't like them it's easy to remove them entirely, see commit https://github.com/xavierleroy/CompCert/commit/e68cbfc39f318c01d6b826c4812ba656b5b279bf .  Just erase a few lines; no need to write new proofs.

This said I agree with @scuellar that CompCert should not increase permissions during compilation, except to allocate scratch stack memory that corresponds to nothing in the source program.  For this reason we're going to keep `perm_inv` in the master branch.

PR #364, opened 2020-07-13
Fix missing semicolon in Makefile

liyishuai:


bschommer:
Should be fixed by commit 72f7830.

Issue #363, opened 2020-07-05
Menhir sometimes suggests a wrong menhirlib location; defend against it?


xuanruiqi:
Sometimes, `menhir --suggest-menhirLib` returns a wrong location. For example, in the Arch Linux package, we use `dune install --prefix "${pkgdir}/usr" --libdir "${pkgdir}$(ocamlfind printconf destdir)"` to install Menhir, but the location suggested by `menhir --suggest-menhirLib` is wrong! It suggests `/usr/lib/menhirLib`, which does not even exist.

However, `ocamlfind` can always return the correct location for Menhir, so maybe a defensive test can test for a faulty `--suggest-menhirLib` and fallback to `ocamlfind`?

jhjourdan:
Cc @fpottier 

Could you please report this bug in Menhir's bugtracker (https://gitlab.inria.fr/fpottier/menhir/-/issues).

xuanruiqi:
I can't create issues in the GitLab, so I have to wait for someone with access to GitLab to report this.

fpottier:
Thanks for reporting this. If you are willing to depend on `ocamlfind`, then you should indeed use `ocamlfind query menhirLib`.

Menhir's `--suggest-menhirLib` dates back to the days when `ocamlfind` was not universally adopted. Its implementation today is not satisfactory (it makes a wild guess, which can be wrong). Maybe I should add a dependency of Menhir on `ocamlfind`, and re-implement `--suggest-menhirLib` to just call `ocamlfind query menhirLib`.

https://gitlab.inria.fr/fpottier/menhir/-/issues/42

xavierleroy:
I fixed this in CompCert's configure script by calling `ocamlfind query menhirLib` first, and only if it fails, calling `menhir --suggest-menhirLib`.  

I agree the way `--suggest-menhirLib` is implemented in Menhir today is not satisfactory.  Earlier versions of Menhir tried harder to report the actual installation directory instead of guessing.

tavianator:
This broke the build for me, since ocamlfind doesn't find menhir:

```
$ ocamlfind query menhirLib
ocamlfind: Package `menhirLib' not found
```

and the logic to fall back to `menhir --suggest-menhirLib` doesn't work because the exit status of a pipeline is that of the last command in it, which is a `tr` that succeeds even when `ocamlfind` fails:

```
$ ocamlfind query menhirLib
ocamlfind: Package `menhirLib' not found
$ echo $?
2
[...]

PR #365, opened 2020-07-13
Revised detection of menhirLib directory

xavierleroy:
This is a follow-up to commit 3b1f3dd5, which was wrong in that errors in a shell pipeline were not correctly detected, as reported
in https://github.com/AbsInt/CompCert/issues/363#issuecomment-657234764

Fixes: #363

### DIFF
--- a/configure
+++ b/configure
@@ -592,11 +592,10 @@ case "$menhir_ver" in
   20[0-9][0-9][0-9][0-9][0-9][0-9])
       if test "$menhir_ver" -ge $MENHIR_REQUIRED; then
           echo "version $menhir_ver -- good!"
-          menhir_dir=$(ocamlfind query menhirLib 2>/dev/null \
-                           | tr -d '\r' | tr '\\' '/') || \
-          menhir_dir=$(menhir --suggest-menhirLib \
-                           | tr -d '\r' | tr '\\' '/') || \
+          menhir_dir=$(ocamlfind query menhirLib 2>/dev/null) || \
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -592,11 +592,10 @@ case "$menhir_ver" in
   20[0-9][0-9][0-9][0-9][0-9][0-9])
       if test "$menhir_ver" -ge $MENHIR_REQUIRED; then
           echo "version $menhir_ver -- good!"
-          menhir_dir=$(ocamlfind query menhirLib 2>/dev/null \
-                           | tr -d '\r' | tr '\\' '/') || \
-          menhir_dir=$(menhir --suggest-menhirLib \
-                           | tr -d '\r' | tr '\\' '/') || \
+          menhir_dir=$(ocamlfind query menhirLib 2>/dev/null) || \
+          menhir_dir=$(menhir --suggest-menhirLib) || \
```
bschommer, line 596: One could also redirect stderr for the second command.

xavierleroy: I thought about it, but the two commands are a bit different: it is normal that `ocamlfind query menhirLib` fails (if Menhir wasn't installed through ocamlfind), so we don't want to show the error message; it is not normal that `menhir --suggest-menhirLib` fails, that's indicative of a problem in Menhir itself, so maybe we want to see the full error message to help track down the problem.

xavierleroy:
Thanks for the quick review.  Now merging.

configure
bschommer: Makes sense.

PR #368, opened 2020-09-08
Upgrade to Flocq 4.0.

silene:
This pull request is not meant to be merged, if only because there is no Flocq 4.0 yet. It was more of an experiment to see how much compatibility was broken by the upgrade of Flocq.

Since Flocq 3.3, there is a formalization of floating-point numbers with a single NaN (or, equivalently, many indistinguishable ones) provided by the file `BinarySingleNaN.v`. Flocq 4.0 reduces the duplication between `BinarySingleNaN.v` and `Binary.v` by making the latter reuse the theorems of the former. This causes a bit of noise in CompCert's files; the function that suffers the most from it is `IEEE754_extra.Bparse`.

Note that, contrarily to Flocq proper, CompCert's embedded version only ships the compatibility version of `SpecFloatCompat.v`, which means that it cannot be used to perform hardware floating-point computations.

(Edit: changed 3.4 to 4.0.)


xavierleroy:
Thanks for the heads up and early updates.  `Bparse` looks even scarier than before, but otherwise the changes to the CompCert files look quite small.

> Note that, contrarily to Flocq proper, CompCert's embedded version only ships the compatibility version of SpecFloatCompat.v, which means that it cannot be used to perform hardware floating-point computations.

So far CompCert doesn't do much compile-time FP arithmetic, so using hardware FP operations is not really needed.  (But the Verasco static analyzer did use hardware FP in the end, I think.)

Will we still be able to use a separately-installed Flocq instead of the vendored Flocq with minimal editing?  or, once I get to work on the configure script again, with no editing and just a configure-time choice?

silene:
> `Bparse` looks even scarier than before

Yes, I am considering adding a few definitions/lemmas to Flocq, so that `Bparse` reverts back to a simpler version.

> Will we still be able to use a separately-installed Flocq instead of the vendored Flocq with minimal editing?

I cannot think of anything that could have broken it. Still, I will make sure it still works for the final version of the pull request.

> So far CompCert doesn't do much compile-time FP arithmetic, so using hardware FP operations is not really needed.

To clarify, this restriction is just to simplify the integration. (No need to make CompCert's build system more complicated if there is no use case yet.) It will be lifted once the minimal supported version is Coq 8.11, or when CompCert is built using an external Flocq.

silene:
As explained in #383, the changes in this pull request did not make it into Flocq 3.4 and are delayed to the next major release, i.e., Flocq 4.0. This pull request will still be useful (once rebased on top of #383), as it contains the changes needed to eventually support Flocq 4.0.

silene:
This pull request has now been updated so as to use Flocq 4.0.

This makes Coq 8.12.0 the minimally supported version. This could be weakened to 8.11.0 if `BinarySingleNaN.Bleb_correct` was commented out, as it is unprovable with Coq 8.11. Earlier versions of Coq cannot be supported without causing too large a divergence between the vendored Flocq and the released ones.

xavierleroy:
I rebased on top of the current master.  The Coq code builds fine, but I'm running into an extraction problem:
```
Warning: The extraction is currently set to bypass opacity, the following
opaque constant bodies have been accessed
: ConstructiveCauchyReals.CRealLt_lpo_dec
  ConstructiveRcomplete.CRealLtDisjunctEpsilon Rdefinitions.Rrepr_appart_0
  Rdefinitions.Req_appart_dec
  ConstructiveExtra.constructive_indefinite_ground_description_Z
  ClassicalDedekindReals.DRealQlim_rec ClassicalDedekindReals.DRealQlimExp2
  ClassicalDedekindReals.DRealQlim ClassicalDedekindReals.lowerCutBelow
  ClassicalDedekindReals.lowerCutAbove solve_constraints_terminate
  Rdefinitions.total_order_T.
 [extraction-opaque-accessed,extraction]
Warning: The following axiom must be realized in the extracted
code: ClassicalDedekindReals.sig_forall_dec.
 [extraction-axiom-to-realize,extraction]
```
Files `extraction/Rdefinition.{ml,mli}` are created, dragging in the unwanted axiom, yet there's no use of Rdefinition in any of the generated OCaml files.  Do you have an idea which dependency to R needs to be cut, on top of those already listed in `extraction/extraction.v`?  I spent some time grepping around but could not figure it out...

silene:
Adding the following commands seems to fix the issue.

```coq
Extract Inlined Constant Rdefinitions.R0 => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.R1 => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.RbaseSymbolsImpl.Rabst => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.RbaseSymbolsImpl.Rrepr => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.Rplus => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.Rmult => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.Ropp => "fun _ -> assert false".
Extract Inlined Constant Rdefinitions.Rinv => "fun _ -> assert false".
```

My current theory is that these commands are needed because the corresponding definitions are not at toplevel but inside an actual module, which confuses Coq's extraction mechanism. But if this is a new issue due to this pull request, my theory is presumably wrong. So, I am going to search a bit more whether some definition in Flocq is responsible.

silene:
It seems like the line following line is causing much trouble.
```coq
Module BSN := BinarySingleNaN.
```

Rather than just an alias, the extraction mechanism considers it as a whole new set of definitions. (The full module signature will appear in the `.mli` file.) So, all the symbols are considered for extraction, even those with an `Extract Inlined Constant`. (This is possibly a Coq bug.) Thus, dependencies are no longer cut.

I should experiment with removing this line from Flocq and see whether it improves the situation. The downside (assuming it works) is that CompCert would depend on Flocq 4.1 instead of 4.0, so a slight additional delay.

silene:
I gave it a try and it works well. In fact, it makes the whole block of commands in `extraction.v` pointless, as Coq now understands by itself that there is no dependency on `R`. (See last commit.)

As mentioned above, this approach has the downside that this is (not yet) a released version of Flocq. I don't expect much changes with the future release of Flocq 4.1.0, but I wouldn't be surprised if a few things have to be fixed to please Coq 8.16.

jdrprod:
Hi @silene @xavierleroy,
I am currently working on CompCert at AbsInt and I just started to write some Flocq extensions that depend on the version of Flocq (and the ability of Flocq to be extracted). @bschommer pointed out to me that this merge request was in progress and I wonder whether I should wait for this PR to be merged or if I can submit my code on top of this PR to avoid having to rewrite everything afterward. I'm asking because I know for sure that my code is hard to integrate with the last version of Flocq (I already tried and I ran into troubles due to the extraction issues already mentioned above).

I could also submit my code as a PR to Flocq (it's essentially a small set of conversion functions and pretty printers for floating points numbers) or even wait for this PR to be merged if does not take long.

andrew-appel:
Hello everyone.  When you are ready, please let me know when you consider this PR as "mature enough" that I should start using it as the basis for porting VST to Flocq 4.0 or 4.1.  Basically I have the same question as @jdrprod , should I wait for this PR to be merged?

silene:
As mentioned above, Flocq is currently not compatible with Coq 8.16 and that is something I want to fix for the release. So, I need to allocate a few hours to complete this task. I should be able to do so by the end of the week, hopefully.

xavierleroy:
Don't rush!  If you can do it within a month, that should be fine for CompCert's next release.  I also have to make sure it works with Coq 8.16 too.

silene:
Flocq 4.1.0 has now been released and this pull request is up-to-date. I have also verified that `-use-external-Flocq` works fine. The pull request was tested with Coq 8.15.1 only, so I do not promise anything regarding the compatibility of `lib/IEEE754_extra.v`. (But I do not expect any kind of trouble.)

xavierleroy:
Excellent, thanks!  I'm reconfiguring the CompCert CI to remove Coq versions < 8.12 and focus on more recent versions.  Then I'll be able to run this PR against several versions of Coq.  Stay tuned.

xavierleroy:
Compiles fine with Coq 8.12.0, 8.14.1 and 8.15.1, on all 5 architectures supported by CompCert.  I'll merge tomorrow.

PR #369, opened 2020-09-18
Simplify two scripts in Zbits

maximedenes:
Previous scripts where relying on the order in which `apply`'s HO
unification performs reductions, for a goal that could be solved by
`reflexivity`.

For more context, I discovered this while working with Matthieu on porting `apply` to the "evarconv" unification engine (https://github.com/coq/coq/pull/991).

The patch has been tested on Coq's master.

### DIFF
--- a/lib/Zbits.v
+++ b/lib/Zbits.v
@@ -266,7 +266,7 @@ Qed.
 Remark Ztestbit_shiftin_base:
   forall b x, Z.testbit (Zshiftin b x) 0 = b.
 Proof.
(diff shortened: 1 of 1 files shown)


xavierleroy:
Thanks for the hint.  It's going through CI right now, to test it with older Coq versions.

xavierleroy:
Looks like it works fine as far bas as Coq 8.8.2.  Merging!

maximedenes:
Thanks!

PR #371, opened 2020-09-22
Refactoring in RTLgen

Artalik:
Using the monad defined above in RTLgen and adjusting some proofs


xavierleroy:
I forgot about this PR, apologies.  But I agree it's nicer to use the state and error monad for `reserve_labels`, even though it can never fail.  Manually merged this PR in master, commit b816d6967 .

PR #380, opened 2020-12-21
Use toolprefix for archiver ar when building libcompcert.a

MSoegtropIMC:
For compiling a cross CompCert, e.g. a 64 bit CompCert on 32 bit Cygwin, it is sufficient to set the toolprefix option. But trying to build an executable with ccomp fails with an error that libcompcert.a doesn't have an index. Using the toolprefix also for the archiver when building libcompcert.a fixes this issue.

This fix is required to build a working 64 bit CompCert for a 32 bit Windows Coq. Some researchers prefer the 32 bit version of Coq even on 64 bit Windows because it is faster (needs less memory and memoy bandwidth).

Please note that I plan to include this change in the CompCert 3.8 opam packages.

### DIFF
--- a/configure
+++ b/configure
@@ -237,6 +237,7 @@ if test "$arch" = "arm"; then
   clinker="${toolprefix}gcc"
   cprepro="${toolprefix}gcc"
   cprepro_options="-std=c99 -U__GNUC__ '-D__REDIRECT(name,proto,alias)=name proto' '-D__REDIRECT_NTH(name,proto,alias)=name proto' -E"
+  archiver="${toolprefix}ar"
   libmath="-lm"
   system="linux"
 fi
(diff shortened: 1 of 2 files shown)


xavierleroy:
Well spotted, thank you for reporting this issue and for the proposed fix.

I'll fix it slightly differently so that the correct archiver is selected also when the Diab toolchain is used.  (The archiver is called `dar` and has different command-line options than `ar`...)

Testing is in progress.

MSoegtropIMC:
Perfect, thanks! I hope it is agreeable to include your fix as patch in the CompCert 3.8 opam package. Since this shouldn't change anything on platforms where it does work without this, I would still call it version 3.8.

xavierleroy:
See commit ca2ebae01 .  You are most welcome to use it as a patch for the CompCert OPAM package.

PR #382, opened 2020-12-27
allow mandir to be overridden for BSD

didickman:
On OpenBSD man pages are installed under /usr/local/man. This commit allows for an override of the location.

### DIFF
--- a/configure
+++ b/configure
@@ -18,6 +18,7 @@
 prefix='/usr/local'
 bindir='$(PREFIX)/bin'
 libdir='$(PREFIX)/lib/compcert'
+mandir='$(PREFIX)/share/man'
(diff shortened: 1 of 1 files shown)


xavierleroy (approved):
Thanks for the patch, this looks very good!  

It's hard to know which of `/usr/local/share/man` and `/usr/local/man` should be the default, but let's keep `share/man` as the default for backward compatibility.

MSoegtropIMC:
Please note that there also is/was a bug in the opam files for CompCert 3.8 - it simply should configure -prefix in opam. This is fixed with https://github.com/coq/opam-coq-archive/pull/1543 (not yet merged).

PR #383, opened 2020-12-28
Update Flocq to 3.4.0.

silene:
This pull request upgrades the embedded Flocq from version 3.1.0 to version 3.4.0, which is about to be released. There are no changes outside the `flocq` directory, which means that the lower bound on Flocq can be left unchanged. By the way, compilation using an external Flocq has also been tested.

The main advantage for CompCert is that `omega` is no longer used in this new version, which considerably reduces the amount of deprecation warning with newer versions of Coq.

This pull request temporarily supersedes #368. The changes in #368 felt a bit too disruptive for a minor release of Flocq, so they are delayed until Flocq 4.0. Pull request #368 is not completely obsolete, though, as it contains changes outside the `flocq` directory, which will eventually be needed, once Flocq 4.0 materializes.

The same disclaimer as in #368 applies. Contrarily to Flocq proper, CompCert's embedded version only ships the compatibility version of `SpecFloatCompat.v`, which means that it cannot be used to perform hardware floating-point computations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ GPATH=$(DIRS)
 
 ifeq ($(LIBRARY_FLOCQ),local)
 FLOCQ=\
+  SpecFloatCompat.v \
   Raux.v Zaux.v Defs.v Digits.v Float_prop.v FIX.v FLT.v FLX.v FTZ.v \
   Generic_fmt.v Round_pred.v Round_NE.v Ulp.v Core.v \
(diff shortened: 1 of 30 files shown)


xavierleroy (approved):
Looks very good, thank you!  

I ran the CI test that checks for compatibility with several versions of Coq (starting with 8.8), and everything looks fine.

Now I can restart my work on replacing `omega` with `lia` elsewhere in CompCert.

PR #394, opened 2021-04-27
Minimal support for `__builtin_unreachable` and for `__builtin_expect`

xavierleroy:
These two builtins are used by GCC and Clang as optimization hints.  They occur frequently in open-source code.

This PR adds minimal support for both builtins.  They are not taken into account by optimizations yet.
-  `__builtin_unreachable` is transported all the way to assembly, where it generates no code.  In the future, it could be used during the Linearize pass to remove more unreachable code.
- `__builtin_expect` is currently removed during C2C elaboration, so it does not even enter the verified part of CompCert.  If we keep it as a "no op" builtin, inefficient code is generated for comparisons on 64-bit platforms.  (Because `__builtin_expect` forces a conversion of its argument to type `long`.)  More work is needed to take full advantage of this builtin.

### DIFF
--- a/aarch64/Asmexpand.ml
+++ b/aarch64/Asmexpand.ml
@@ -355,8 +355,12 @@ let expand_builtin_inline name args res =
   (* Synchronization *)
   | "__builtin_membar", [], _ ->
      ()
+  (* No operation *)
   | "__builtin_nop", [], _ ->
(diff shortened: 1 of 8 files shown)


xavierleroy:
Positive feedback received, merging!

Issue #395, opened 2021-04-30
Suspected bug in elaboration phase supporting `-fbitfields` introducing sequence point violation


tduckwo:
Consider the following C source, `bug.c`. I believe it should result in the
printing of 0, but with CompCert it prints 1.

```c
int printf(char *, ...);

struct {
  int m1 : 4;
  int m2 : 4;
} volatile s = {1, 1};

int f() {
  s.m2 = 0;
  return 0;
}

int main() {
  s.m1 = f();
  printf("%d\n", s.m2);
  return 0;
}
```

Based on my understanding of C99, there's a trivial sequencing of the evaluation
of `f()` before the assignment of its result to `s.m1` (which, to my knowledge,
does not imply any read of the previous value of `s.m1` or `s.m2` except perhaps
during the execution of `f`). This implies that `s.m2` is set to `0` during the
evaluation of `f()`, and that `s.m1` is set to `0` thereafter as part of the
assignment in `main`. Thus, we expect that the read of `s.m2` as part of the
call to `printf` should return `0`, and that `0` should be printed to standard
out. Using CompCert 3.8 compiled for `x86_64-linux`, we may observe the
following (this also appears to hold for a newer revision,
`e4542668e6d348e0300e76bb77105af24aff4233`).

```bash
$ ccomp -fbitfields -interp -quiet bug.c
1

$ ccomp -fbitfields -o bug bug.c

$ ./bug
1

$ gcc -o nobug bug.c

$ ./nobug
0

$ ccomp -fbitfields -interp -quiet -random bug.c
0
```

Using `-dparse`, we arrive at the following elaborated program.

```c
extern void __builtin_debug(int kind, ...);

int printf(char * , ...);

struct _139;

struct _139 {
  unsigned char __bf1;
};

struct _139 volatile s =
  {(unsigned char) ((1 & 0xFU) << 0 | (1 & 0xFU) << 4), };

int f(void)
{
  s.__bf1 = s.__bf1 & ~0xF0U | (unsigned int) 0 << 4U & 0xF0U;
  return 0;
}

int main(void)
{
  s.__bf1 = s.__bf1 & ~0xFU | (unsigned int) f() << 0U & 0xFU;
  printf("%d\n", (int) (s.__bf1 << 24) >> 28);
  return 0;
  return 0;
}
```

The bitfield elaboration appears to introduce a sequence point violation.
Depending on the order in which `s.__bf1` is read and written (read on the
[...]

PR #400, opened 2021-06-24
Native support for bit fields

xavierleroy:
This big PR adds support for bit fields in structs and unions to the verified part of CompCert, namely the CompCert C and Clight languages.

The compilation of bit field accesses to normal integer accesses + shifts and masks is done and proved correct as part of the Cshmgen pass.

The layout of bit fields in memory is done by the functions in module Ctypes.  It follows the ELF ABI layout algorithm.  As a bonus, basic soundness properties of the layout are shown, such as "two different bit fields do not overlap" or "a bit field and a regular field do not overlap".

All this replaces the previous emulation of bit fields by source-to-source rewriting in the unverified front-end of CompCert (module cparse/Bitfield.ml). This emulation was prone to errors (see #395) and could not express the ELF ABI layout, implementing its own nonstandard layout instead.

The core idea for the PR is that expressions in l-value position denote not just a block, a byte offset and a type, but also a bitfield designator saying whether all the bits of the type are accessed (designator `Full`) or only some of its bits (designator `Bits`).  Designators of the `Bits` kind appear when the l-value is a bit field access; the bit width and bit offset in `Bits` are computed by the functions in `Ctypes` that implement the layout algorithm.

Consequently, both in the semantics of CompCert C and Clight and in the SimplExpr, SimplLocals and Cshmgen compilation passes, pairs of a type and a bitfield  designator are used in a number of places where a single type was used before.

The introduction of bit fields has a big impact on static initialization (module cfrontend/Initializers.v), which had to be rewritten in large part, along with its soundness proof (cfrontend/Initializersproof.v).

Both static initialization and run-time manipulation of bit fields are tested in `test/abi` using differential testing against GCC and randomly-generated structs.

This work exposed subtle interactions between bit fields and the `volatile` modifier.  Currently, the `volatile` modifier is ignored when accessing a bit field (and a warning is printed at compile-time), just like it is ignored when accessing a struct or union as a r-value.  The alternative -- emitting events in the trace when accessing a volatile bit-field -- raises many semantic difficulties that I was not able to overcome.

Currently, the natural alignment of bit fields and their storage units cannot be modified with the `aligned` attribute.  `_Alignas` on bit fields is rejected as per C11, and the `packed` modifier cannot be applied to a struct containing bit fields.  It might be possible to support bit fields in packed structs if required.


xavierleroy:
Manually squashed and merged into master.

PR #402, opened 2021-07-16
Support -MD and -MMD options and pass through to the preprocessor

moehriegitt:
Hi!

This adds support for -MD and -MMD pass-through to the pre-processor.

**Henrik


moehriegitt:
Shoot, it should be  'Self gnu_prepro_opt', not  'String (gnu_prepro_opt_key "-MD")'. Please reject this request, I'll add a new one.

didickman:
Although this pull request was closed (and I do not see a new pull request), I would add that this feature would be useful for me too.

PR #404, opened 2021-07-30
Add csyntaxgen tool for exporting CompCert C (Csyntax) AST as .v file

btj:
Adds the `csyntaxgen` tool, which takes a `.c` file and exports the CompCert C AST (as defined in `Csyntax.v`) to a `.v` file.

Produced by copy-pasting and adapting the `exportclight` and `tests/clightgen` directories. I left these steps as separate commits to ease reviewing; of course, these should be squashed.

## Motivation

The Clight AST, as exported by the `clightgen` tool, is most appropriate when a human user needs to directly interact with the AST, such as when using the Verified Software Toolchain. For certain other use cases, however, the CompCert C AST may be more appropriate.

Specifically, we want to prove soundness of [CH2O](https://robbertkrebbers.nl/research/ch2o/) with respect to CompCert C, and then obtain theorems about CompCert-compiled assembly code by first verifying the `.c` program using our [VeriFast](https://github.com/verifast/verifast) verification tool, which will eventually emit a Coq proof of correctness with respect to CH2O, and then porting this correctness statement to CompCert C using the soundness theorem, and then applying the CompCert correctness proof.

The CH2O-CompCert soundness theorem will be based on a relation linking a CH2O Core C program to a corresponding CompCert C program. This relation is easier to define between CH2O Core C and CompCert C than between CH2O Core C and Clight, and easier to prove for the CH2O Core C AST generated from a .c file and the CompCert C AST generated from that .c file, because CompCert C is closer to the source code (and closer to CH2O Core C) than Clight.

In any case, we want to prove soundness of CH2O with respect to CompCert C, not Clight, because the former is a stronger statement.


xavierleroy:
Thank you @btj for this code!  I agree it's good to have tools to support working with the CompCert C AST.  I'll review and get back to you later this month.

xavierleroy:
PR #413 takes this PR and shares more of the code with clightgen.  #413 was merged, so I'm closing this PR.  Thanks for the suggestion and for the code!

Issue #403, opened 2021-07-29
Global considered harmful


andrew-appel:
Since an important application of CompCert (especially the dual-licensed portion) is to be used as a library in other projects, it would be polite to avoid "Global" declarations in CompCert -- these interfere with client projects even if the client just does `Require` without `Import`.

Especially `Global Set Asymmetric Patterns` in Coqlib.v.  I conjecture one could safely remove `Global` here, because (within CompCert) almost every `Require` of Coqlib is also `Import`.

Generally the `Global Hint Resolve` declarations are not _too_ bad, because _most_ of those are in hint databases specific to compcert (and therefore will not interfere much with clients).  However, it would still be good form to turn all of those into` #[export] Hint Resolve` if possible.

And generally the `Global Opaque` all seem to control definitions created within CompCert, so these are probably not harmful.

xavierleroy:
> Especially Global Set Asymmetric Patterns in Coqlib.v. I conjecture one could safely remove Global here, because (within CompCert) almost every Require of Coqlib is also Import.

No, non-Global `Set Asymmetric Patterns` applies to the current file only, not to files that import it.  

This `Global Set Asymmetric Patterns` was a quick and irritated reaction to a sudden, backward-incompatible, never publically discussed change in Coq 8.5 (https://coq.github.io/doc/master/refman/changes.html#id984).  You, too, would have been irritated if someone changed the semantics of pattern matching in SML without telling you in advance.

This said, I agree `Global Set Asymmetric Patterns` is not nice.  See #408 for an attempt to get rid of it.

> Generally the Global Hint Resolve declarations are not too bad, because most of those are in hint databases specific to compcert 

These declarations follow Dr. Pédrot's elaborate migration plan (https://coq.discourse.group/t/change-of-default-locality-for-hint-commands-in-coq-8-13/1140).  

> However, it would still be good form to turn all of those into #[export] Hint Resolve if possible.

This requires going over all CompCert proofs and fixing obscure breakage.  I wish Coq would provide tools to help with this.

> And generally the Global Opaque all seem to control definitions created within CompCert, so these are probably not harmful.

There is no other way (that I know of) to define something and make it opaque to `simpl`, `cbn, etc, while not making it fully opaque as in `Proof... Qed.`.  An attribute over `Definition` would be nicer, but Coq doesn't seem to support this.

PR #408, opened 2021-08-22
Avoid `Global Set Asymmetric Patterns`

xavierleroy:
Instead, add `Set Asymmetric Patterns` to the files that need it,
or use `Arguments` to make inductive types work better with symmetric patterns.

Closes: #403


xavierleroy:
I'll review myself since nobody does it.  I have a good feeling about this change.  Few files really need `Set Asymmetric Patterns`, and a couple of files look better now that they've been converted to symmetric patterns.  Good to merge!

PR #415, opened 2021-09-26
Improved compatibility with Coq 8.14

xavierleroy:
As usual, version 8.14 of Coq introduces new warnings, some of which cannot be honored without breaking backward compatibility with 8.13 and earlier.  (Grmph.)

This PR is a best effort: address the warnings that can be addressed, and silence the others.

Backward compatibility was successfully tested with Coq 8.13, 8.11, and 8.9.


xavierleroy:
Coq 8.14.0 is about to be released, so it's time to merge this PR.

PR #416, opened 2021-10-01
Synchronize with vendored MenhirLib

jhjourdan:
A diff between Compcert's vendored coq-menhirlib and upstream revealed some differences. Most of them correspond to what is covered in #415, but there are also a few other changes.

This should not break compatibility with the versions of Menhir which are already compatible (i.e., from 20190626).

### DIFF
--- a/MenhirLib/Alphabet.v
+++ b/MenhirLib/Alphabet.v
@@ -37,7 +37,7 @@ Qed.
 Definition comparableLt {A:Type} (C: Comparable A) : relation A :=
   fun x y => compare x y = Lt.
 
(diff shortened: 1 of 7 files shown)


xavierleroy:
Thanks!  You're right that I didn't pick the `Fail_sr` -> `Fail_sr_full` recent change yet, but it's probably time to do it, as compatibility with earlier Menhir versions seems unaffected.  To be continued...

xavierleroy (approved):
Looks good to me, and CI on various Coq versions is happy.  Thanks!

PR #417, opened 2021-10-01
Adapt to coq/coq#13837 ("apply with" does not rename arguments)

SkySkimmer:


xavierleroy (approved):
Thanks for the advanced warning.  The change seems compatible with Coq 8.9 to 8.13.  Merging!

PR #425, opened 2022-01-06
Adapt w.r.t. coq/coq#15442.

ppedrot:

### DIFF
--- a/common/Events.v
+++ b/common/Events.v
@@ -198,7 +198,7 @@ Next Obligation.
   elimtype False. elim NE. auto.
 Qed.
 Next Obligation.
-  red; intro. elim (H e). rewrite H0. auto.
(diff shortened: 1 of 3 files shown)


xavierleroy:
Thanks for the early warning and the patch.

I rarely use "match goal with" in my scripts (outside of Ltac macros).  Nothing wrong with it, just a question of habit.

I pushed some further tweaks on top of yours that should also succeed in removing dependencies on generated names in the proofs in question.

ppedrot:
This will have to be backported to the compcert bundled with VST then, [the corresponding PR](https://github.com/PrincetonUniversity/VST/pull/536) has been merged already.

xavierleroy:
I checked with https://github.com/coq/coq/pull/15442 and everything seems right.  Merging!

PR #427, opened 2022-01-19
Fix a typo in a comment in Clight.v

MisakaCenter:


xavierleroy:
Please don't close!  It will be addressed, just with low priority.

PR #429, opened 2022-05-02
Extends boolean_equality to constructor with more arguments and lists

jjhugues:
This PR extends boolean_equality to constructor with more arguments and lists.
It has been tested on my own development to work with Coq 15.1


xavierleroy (approved):
Looks good to me.  The `boolean_equality` tactic is a terrible hack overall, but this doesn't make it any worse...

Issue #436, opened 2022-06-21
Jump macro instructions in the RISC-V target printer


Yuki21:
I found a relocation error while compiling heavy programs on RISC-V:
Using CompCert, if the destination symbol of a goto *or* a tailcall (`Mtailcall sig (inr s)` in Mach) is far away from the current pc (for program size over 1MB), then the address may not fit in the 21-bits address offset supported by the `jal` (jump and link) instruction.
In such a case, the linker will reject the program (fortunately, no buggy code is produced).

To fix this problem, we can modify the target printer and rely on the linker relaxation:

As `Mgoto` and `Mtailcall` are translated to `Pj_l` and `Pj_s` respectively, we just have to modify the following code in the target printer (replacing the `j` pseudo instruction by a `tail`:

from
```ocaml
      | Pj_l(l) ->
         fprintf oc "	j	%a\n" print_label l
      | Pj_s(s, sg) ->
         fprintf oc "	j	%a\n" symbol s
```

to
```ocaml
      | Pj_l(l) ->
         fprintf oc "	tail	%a\n" print_label l
      | Pj_s(s, sg) ->
         fprintf oc "	tail	%a\n" symbol s
```

Indeed, `j` is a pseudo instruction corresponding to a `jal x0, offset` while `tail` is expanded into the sequence:
```
auipc x6, offset[31 : 12] + offset[11]
jalr x0, offset[11:0](x6)
```

It is not a problem to always generate `tail` pseudo instructions instead of `j` since the linker is able to replace the `tail` pair to a single `jal` when the address fits.

For reference and discussion on this topic, see the following resources:
- Pseudo instructions for jumps on RISC-V: [Volume I Unprivileged ISA V20191213, p140](https://github.com/riscv/riscv-isa-manual/releases/download/Ratified-IMAFDQC/riscv-spec-20191213.pdf)
- SiFive forum discussion [here](https://forums.sifive.com/t/link-error-relocation-truncated-to-fit-r-riscv-jal-against-symbol/4163)
- SiFive blog article about linker relaxation on RISC-V [here](https://www.sifive.com/blog/all-aboard-part-3-linker-relaxation-in-riscv-toolchain)

Is the target printer written this way on purpose, or is this a bug? We consider it is a bug on our side, and we have fixed our CompCert fork as explained above.

xavierleroy:
[...]

PR #437, opened 2022-06-22
proposed bugfix for issue #436

Yuki21:
More information about the bugfix:
1. We don't need to modify the TargetPrinter for `Pj_l`, as it is only used for goto inside a function. (We can suppose that the case where an address does not fit is very unlikely for gotos inside a function; moreover, even if we have a huge function where it happens, we would then have more important problems to deal with such as the conditional jumps which are addressed on 12bits only).
2. Of course, modifying the `Pj_s` printer rule as above implies a modification in the Asm semantics of CompCert, as the pseudo instruction clobbers x6. This modification itself implies some small changes in the proof, and the need to declare x6 as a scratch register. It must thus be invisible from Mach.

I modified the Asm semantics, as well as the Machregs builtins parameters, because x6 can not be used as an input for builtins anymore. Thus, we use x8 instead, and, consequentially, I adapted Asmexpand untrusted file to change builtins functions and corresponding assertions.

Don't hesitate to tell me if I forgot something here, but I think this commit solves the issue. The non-regression tests are passing successfully.

### DIFF
--- a/riscV/Asm.v
+++ b/riscV/Asm.v
@@ -751,7 +751,9 @@ Definition exec_instr (f: function) (i: instruction) (rs: regset) (m: mem) : out
   | Pj_l l =>
       goto_label f l rs m
   | Pj_s s sg =>
-      Next (rs#PC <- (Genv.symbol_address ge s Ptrofs.zero)) m
+      Next (rs#PC <- (Genv.symbol_address ge s Ptrofs.zero)
+              #X6 <- Vundef
+           ) m
   | Pj_r r sg =>
       Next (rs#PC <- (rs#r)) m
   | Pjal_s s sg =>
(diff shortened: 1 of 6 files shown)


Yuki21:
Your comment made me want to dig a little deeper into the question.
The `call` macro is using x1 (the return address) instead of x6. I thought the only way to avoid loosing x6 was to generate manually the auipc + jalr pair, but in fact it is not a great solution as you will not succeed to obtain the right relocation tag.
Loading the relative address manually using:
```
auipc	a0, %pcrel_hi(msg + 1)
addi	a0, a0, %pcrel_lo(1b)
```
will generates relative address loading tags in the object file, and it is not exactly what we want.

However, there is a special pseudo instruction `jump` (see https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#function-calls) that can do the job. This pseudo instruction is expanded in the same way as the `tail` but with a specified scratch register.

For instance, if I have a `jump func, x31` in the asm file: I compile it with the gcc assembler, and then l look the objdump with `riscv64-linux-gnu-objdump -Mnumeric -r -d`. We obtain:
```
  38:   00000f97                auipc   x31,0x0
                        38: R_RISCV_CALL        func
                        38: R_RISCV_RELAX       *ABS*
  3c:   000f8067                jr      x31 # 38 <_start+0x18>
```
and thus we can use x31 as a scratch register for tailcalls (and x31 is already marked as scratch in CompCert)!

I have made a new proposition in https://github.com/Yuki21/CompCert/commit/b14d831ec89f5451fd72f338369e54873045cc34.

xavierleroy:
Thanks for the suggestion to use `jump symb, x31`.  The `jump` pseudo is not documented in the 2019 RISC-V ISA manual, but has been mentioned in the assembly language manual since 2020, and the GNU assembler handles it fine.  I didn't see your second push before implementing the `jump`-based solution.  It's now in trunk.  Let me know if it doesn't work for you.

PR #439, opened 2022-06-30
Adapt w.r.t. coq/coq#16004.

ppedrot:
The change should be compatible up to 8.11, but since CompCert only guarantees compatibility up to 8.12 this is fine.


xavierleroy:
Thanks for the patch.  A bit of CI confirms that this works fine in 8.12 and up.

In a follow-up commit, I'll re-enable the deprecated-hint-rewrite-without-locality warning, which should now not trigger any more.

PR #443, opened 2022-07-18
improved auto goal selection

mrhaandi:
Improves robustness in case of stronger (e)auto (see [coq/#16293](https://github.com/coq/coq/pull/16293)).

### DIFF
--- a/backend/Lineartyping.v
+++ b/backend/Lineartyping.v
@@ -305,7 +305,7 @@ Local Opaque mreg_type.
 - (* getstack *)
   simpl in *; InvBooleans.
   econstructor; eauto.
-  eapply wt_setreg; eauto. eapply Val.has_subtype; eauto. apply WTRS.
+  eapply wt_setreg; eauto. eapply Val.has_subtype; [eauto|apply WTRS].
   apply wt_undef_regs; auto.
 - (* setstack *)
   simpl in *; InvBooleans.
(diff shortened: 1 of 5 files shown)


mrhaandi:
@xavierleroy I did check that this PR compiles both before and after https://github.com/coq/coq/pull/16293.
The main reason for the proposed changes is that fixing bugs in `auto` may allow it to solve more goals.
Non-monotone use of automation such as `tac1; auto. tac2.` may, therefore, break.

xavierleroy:
@mrhaandi : thanks for the PR and the pointers.  The discussion of the "antipattern" is interesting.  I'd like to seize the opportunity to clean up some of the scripts touched by this PR, so I'll leave it open for a few weeks.  But please go ahead with making (e)auto stronger in Coq, knowing that CompCert will be adapted accordingly.

ppedrot:
@xavierleroy coq/coq#16293 is basically pending on this overlay. It's not strictly urgent but there is always the risk of other devs introducing incompabilities with that PR, so I believe it'd be better to merge this PR quickly.

xavierleroy:
Generally speaking I prefer when Coq is changed and then CompCert is adjusted to the change, rather than doing things in the opposite order.  But if this is really blocking for the Coq change, I'll merge this PR and make a mental note to check again when the Coq changes enter alpha or beta stage.

ppedrot:
Isn't it better to write backwards compatible patches, though? That way no CI is ever broken.

PR #449, opened 2022-08-14
Add support for PowerPC and AArch64 BSD

brad0:


didickman:
Hi Brad, it looks like you copied this from https://github.com/openbsd/ports/blob/master/lang/compcert/patches/patch-configure.

There we also have this line: "For PowerPC targets, the "ppc-" prefix can also be written "powerpc-" and can be refined into:". Do we want to have the same here?

Also did you happen to test your proposal on these respective platforms to make sure they work? A long time ago I was using OpeBSD + CompCert + powerpc, but it's been a few years since I verified this still works.

brad0:
What's there is a copy and paste of what's there swapping Linux for BSD. Modeled after what's there for the existing archs. I asked George to test on PowerPC and he tested on PowerPC and PowerPC64. The TargetPrinter.ml parts for each arch are also necessary. He said the current package on PowerPC is broken without the TargetPrinter.ml patch.

xavierleroy:
After discussion with the team at AbsInt, we prefer not to support these configurations in the sources we publish, as we are unable to test them as part of our CI.  We suggest you continue to apply these patches as part of the build of the OpenBSD package.

PR #450, opened 2022-08-14
Allow sharedir path to be overridden

brad0:


xavierleroy:
This is not going to work: the CompCert executable `ccomp` looks for `compcert.ini` in `$D/../share`, where `$D` is the directory containing `ccomp`.  This makes binary CompCert distributions easy to install and relocate.

brad0:
Ya, I came across the Driver code for that and trying to understand it.

We need to be able to put compcert.ini into a sub-directory of share/ as in share/compcert.

xavierleroy:
I misspoke: CompCert looks for `compcert.ini` in `$D`, $D/../share` and `$D/../share/compcert`.  So your wish could be satisfied, but the configure script needs more work.

xavierleroy:
See proposal at #460.

brad0:
With #460 posted I am going to close this.

PR #456, opened 2022-09-21
Support the -std=<standard> command-line option

xavierleroy:
This PR adds a `-std` command-line option similar to those of GCC and Clang.

Currently three values are supported: `-std=c99`, `-std=c11` and `-std=c18`.

The option has two effects:
- With a GNU toolchain, it's passed down to the preprocessor.  This influences the contents of standard header files.  
- It determines the default handling of the "C11 feature" warning: with `-std=c99`, CompCert warns on the use of C11 features; with `-std=c11` or `-std=c18`, no warnings are emitted.

For backward compatibility with CompCert's current behavior, if no `-std` option is given, the default is to not warn, but `-std=c99` is passed to the preprocessor.  This may change in the future, when C11 will become the default instead of C99.

### DIFF
--- a/driver/CommonOptions.ml
+++ b/driver/CommonOptions.ml
@@ -10,20 +10,21 @@
 (*                                                                     *)
 (* *********************************************************************)
 
+open Printf
 open Clflags
 open Commandline
 
 (* The version string for [tool_name] *)
 let version_string tool_name =
   if Version.buildnr <> "" && Version.tag <> "" && Version.branch <> "" then
-    Printf.sprintf "The CompCert %s, Release: %s, Build: %s, Tag: %s, Branch: %s\n"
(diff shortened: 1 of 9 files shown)


driver/CommonOptions.ml
```diff
@@ -56,6 +70,9 @@ let language_support_options =
 
 let language_support_help =
   {|Language support options (use -fno-<opt> to turn off -f<opt>) :
+  -std=c99  or  -std=c11  or  -std=c18
+                 Choose the ISO C language standard used: C99, C11 or C18.
+                 The "gnu" variants enable some GNU C extensions.
```
bschommer, line 75: The line about the gnu variants can probably be removed, since the option does not support the gnu variants.

```suggestion
```

xavierleroy: Well spotted, thanks!

PR #465, opened 2022-11-22
Adapt to coq/coq#16743 (fix incomplete checking of unsolved holes)

SkySkimmer:
Alternatively we could define `exploit` as a Tactic Notation taking
open_constr or uconstr (the problem is that arguments to a
non-notation Ltac get interpreted as constr which doesn't allow holes,
but in this case a hole gets solved by typeclass inference leaving a
new hole behing, and due to the bug the new hole isn't detected).


xavierleroy:
I'm fine with this change, even though I don't understand a word of the explanations :-)

Why is this submitted as a draft PR ?

SkySkimmer:
Reflex from the usual way we do this

ppedrot:
Please merge now.

PR #470, opened 2022-12-20
Adapt w.r.t. coq/coq#16904.

ppedrot:
This is backwards compatible. I haven't touched to the instances of casetype / elimtype in the flocq subfolder, since these should be removed by the next flocq version bump.


xavierleroy (approved):
It's nicer with `exfalso`, indeed.  Thanks!

PR #481, opened 2023-02-03
for consistency: print flt.s, feq.s, fle.s with a tab after the mnemonic

monniaux:
All RISC-V instructions are printed with a TAB after the mnemonic, except these three single-precision comparisons. I suppose this is an oversight.


xavierleroy (approved):
Makes complete sense.  Thanks for having noticed it.

PR #482, opened 2023-02-14
Improved register allocation for ARM udiv and sdiv.

fajb:
To accommodate hardware with and without support for integer division, the ARM backend imposes strong constraints on register allocation.

For hardware with support for integer division, the PR lifts this restriction.
For hardware without support for integer division, instruction selection generates builtins.

*NB* Though this requires more testing, I think this is worth sharing.

### DIFF
--- a/arm/Archi.v
+++ b/arm/Archi.v
@@ -100,3 +100,6 @@ Parameter abi: abi_kind.
 (** Whether instructions added with Thumb2 are supported. True for ARMv6T2
   and above. *)
 Parameter thumb2_support: bool.
+
+(** Whether the hardware supports sdiv and udiv *)
+Parameter hardware_idiv : unit -> bool.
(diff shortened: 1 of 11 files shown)


xavierleroy:
Apologies for the late reply.  This is an interesting proposal, thanks.  When the ARM port was developed, there was no builtins-with-semantics.  Also, very few ARM chips had integer division instructions.  But it is probably time to reconsider choices made back then.

xavierleroy:
After discussions with @fajb and @bschommer , I realized that the proposed solution falls into a shortcoming of pure builtin functions : currently, they are not eligible for CSE, so the generated code (in the no-hardware-division case) is not as good as what we get today.  Instead, @bschommer implemented 94d2111fa24ddbe8e71f151b985d610c2a8d74f7 to relax the register constraints on division when hardware division is available.

So, allow me to close this PR, after having thanked @fajb again for bringing this issue to our attention.

PR #485, opened 2023-03-19
Fix a comment in Cstrategy.v

Hughshine:
A trivial PR :D


xavierleroy (approved):
Trivial but entirely correct!  Thank you.

PR #487, opened 2023-04-04
Use a shorter insruction on x86 for loading unsigned 32-bit immediates into 64-bit registers.

valoran-M:
Writing into a 32-bit register erases the upper 32 bits of a 64-bit register. This is the shortest instruction for loading unsigned 32-bit immediates. The length of those instructions is:
- movl: 5 bytes
- movq: 7 bytes
- movabsq: 10 bytes

Note: Most assemblers will choose the proper instruction by themselves;  this commit just makes this explicit.

### DIFF
--- a/x86/TargetPrinter.ml
+++ b/x86/TargetPrinter.ml
@@ -432,7 +432,9 @@ module Target(System: SYSTEM):TARGET =
       | Pmovq_ri(rd, n) ->
           let n1 = camlint64_of_coqint n in
           let n2 = Int64.to_int32 n1 in
-          if n1 = Int64.of_int32 n2 then
+          if Int64.shift_right_logical n1 32 = Int64.zero then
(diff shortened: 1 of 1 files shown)


silene:
For the record, this issue showed up when improving the code generation for division by a constant. The algorithm loads a 64-bit immediate between 2^31 and 2^32-1, which was thus emitted as `movabsq` by CompCert. But GCC/GAS honors this choice (to account for self-modifying code?), which means that the machine code would use the 10-byte instruction where the 5-byte one would have been sufficient. This can have adverse effect on instruction decoding and caching.

xavierleroy:
Well spotted!  I was not aware that `movl` produces shorter code than `movq` even in 64 bit mode.   And there's the additional benefit for constants in the 2^31.. 2^32-1 range.  Nice!

xavierleroy (approved):

PR #488, opened 2023-04-15
'spass' test: correctly typed function call

monniaux:
'Spass' is implemented using forms of data abstraction making use of function pointers. Also, some integer values are hidden inside function pointers.

The benchmark, at some point, casts a function pointer for a function taking two integer parameters into a pointer for a function taking two pointer parameters. Then this function is called with pointer parameters.

This seems to me incorrect with respect to the C standard.

In CompCert, if applying inlining after constant propagation, this leads to an ill-typed function.

My proposed fix is to use correctly-typed function pointers.


monniaux:
§6.3.2.3-8 "If a converted pointer is used tocall a function whose type is not compatible with the referenced type, the behavior is undefined." so indeed this was wrong.

xavierleroy:
I guess the original code was written with `sizeof(int) == sizeof(void *)` in mind, in which case it would still not conform to the C standards, but would work with CompCert and other reasonable C compilers.  Thanks for the report and fix.

PR #490, opened 2023-05-05
Refine the abstract domain of values

xavierleroy:
Introduce a `Num p` abstract value that stands for any number (but no pointers), with provenance `p`.  In non-strict mode, this makes the value analysis more precise than using `Ifptr p` for the same purpose, like we did before.   (In particular, it should address the imprecision mentioned in #489, even though the `vlub` function remains non-monotonic.)

Also: minor simplifications and cleanups in the management of provenance. In strict mode, instead of setting all provenances to `Pbot`, just propagate them but ignore them in `aptr_of_aval`.  Simplify `vnormalize` and its proof consequently.


xavierleroy:
NB. This will need to be run against AbsInt's test suite to check that it does not cause over-optimization.

PR #492, opened 2023-05-30
Fix compilation with Coq master

SkySkimmer:
A goal is solved with master but I guess not with whatever version you tested.

See also https://github.com/coq/coq/issues/17660


maximedenes:
How does this relate to #491 ? Is one of the two preferable?

SkySkimmer:
I didn't see #491 
I guess compcert maintainers can pick the one they prefer

ppedrot:
I think the other one is a bit more robust (it doesn't depend on a fixed goal number).

xavierleroy:
Thank you both for the fixes.  I'm about to merge #491 since it looks more robust indeed.

I'm still very surprised that the new "auto" in Coq master can kill one of the goals of this proof entirely, while not breaking anything else in CompCert...

PR #495, opened 2023-06-25
Revert "Do not print cfi_sections for bsd."

brad0:
0af96302ae861e35c7e2dace1182e30be0b67851

FreeBSD / OpenBSD no longer use ancient binutils assembler.


xavierleroy:
Thanks for pointing this out.

PR #503, opened 2023-09-12
To use movt/movw, what matters is not whether the processor is in Thu…

monniaux:
…mb2 mode, but whether the processor supports Thumb2 mode. These instructions are also available in ARM mode on such processors.


xavierleroy:
Well spotted, thanks.

PR #509, opened 2024-01-29
Remove strictly positive size of blocks in free semantics

josuemoreau:
The condition `Ptrofs.unsigned sz > 0` in the semantics of `free` makes it difficult to prove that a program cannot block when calling `free`. So, this commit removes it, at the expense of a slightly more complicated proof that `free` is a proper external function.


xavierleroy:
Thanks for this proposal.  It will take a while before I can review it in depth, but it sounds interesting.

xavierleroy:
I could finally review, and I like the idea very much. Thanks!  I just manually merged this PR, with a slightly simpler proof of the injection diagram in Events.v (commit e1ea590).

PR #518, opened 2024-08-16
CSE: remember the abstract pointers in `Load` equations

xavierleroy:
CSE uses non-aliasing information provided by value analysis to reuse the result of a previous load after a store, provided the load and the store do not alias.

In the current implementation of CSE, the abstract pointer associated with the previous load is not stored in the CSE-inferred equations, but is recomputed at the point of the store just before calling the non-aliasing test `pdisjoint`.  This can lead to loss of precision, as in the following example:
```
    x1 <- addrglobal "x"
    y <- load x1
    y1 <- addrglobal "y"
    store(y1, ...)
    x2 <- addrglobal "x"
    z <- load x2
```
At the point of the store instruction, we no longer know that the value of x1 is a pointer into the "x" global.  (That's because value analysis forgets the abstract values of pseudoregisters when they become dead, so as to avoid memory blowup on large codes.)  Hence, CSE doesn't know that the "store" and the first "load" do not interfere, and forgets the equation describing the first "load".  Consequently, the second "load" cannot be replaced by `z <- y`.

This PR addresses this issue by storing in `Load` equations the abstract pointers describing the location of the read.  These abstract pointers are determined at the time when `Load` equations are generated and when full abstract value information is available.  This produces better code for examples such as the one above, and could also reduce compilation times for CSE, as some recomputations of abstract pointers are avoided.

### DIFF
--- a/backend/CSE.v
+++ b/backend/CSE.v
@@ -77,7 +77,7 @@ Fixpoint find_valnum_rhs (r: rhs) (eqs: list equation)
   match eqs with
   | nil => None
   | Eq v str r' :: eqs1 =>
-      if str && eq_rhs r r' then Some v else find_valnum_rhs r eqs1
+      if str && compat_rhs r r' then Some v else find_valnum_rhs r eqs1
   end.
 
 (** [find_valnum_rhs' rhs eqs] is similar, but also accepts equations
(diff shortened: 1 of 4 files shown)


xavierleroy:
The CI failure on PPC appears unrelated, it occurs also on master but not on my other CI, which uses different versions of qemu.  To be debugged.

monniaux:
@xavierleroy FYI We are also experiencing problems with `qemu-ppc` 7.2.11 in our CI  (`render` produces an illegal instruction in it), which happens neither on `qemu-ppc` 4.2.1 nor on 9.02.

PR #520, opened 2024-08-21
More precise value analysis for "known integer or Vundef" results

xavierleroy:
This PR introduces a new abstract value `IU n`, standing for "either `Vint n` or `Vundef`".

It enables more precise value analysis of comparison and selection operators, which in turn enables more compile-time evaluation of these operators.

Example:
```
struct s { char c[4]; char d[2]; };

int f(void)
{ struct s x; return (x.c + 6 == x.d); }

int g(void)
{ struct s x; if (x.c + 6 == x.d) return 1; else return 0; }
```

The current CompCert optimizes `g` into `return 0`, but fails to optimize `f`, as the result of the pointer comparison can be Vundef, so its abstract value cannot be `I Int.zero`.

With this PR, the expression `x.c + 6 == x.d` is analyzed as `IU Int.zero`, and constant-propagated to the constant 0.  Hence, `f` and `g` produce the same `return 0;` code.

### DIFF
--- a/aarch64/ConstpropOp.vp
+++ b/aarch64/ConstpropOp.vp
@@ -23,7 +23,7 @@ Require SelectOp.
 
 Definition const_for_result (a: aval) : option operation :=
   match a with
-  | I n => Some(Ointconst n)
+  | I n | IU n => Some(Ointconst n)
   | L n => Some(Olongconst n)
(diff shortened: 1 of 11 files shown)


xavierleroy:
Note: we could further reduce `Uns p 0` to `IU Int.zero`, since 0 is the only 0-bit unsigned integer.  But this needs a bit of work on the lub operator.  I'll look into this later.

Issue #526, opened 2024-10-18
`make install` should install `.v` files and `.glob` files in addition to `.vo` files


JasonGross:
If compcert does not install `.v` and `.glob` files, the bug minimizer cannot minimize any bugs in projects that are built on top of compcert.

Note that `coq_makefile`-made makefiles install the .v and .glob files:
https://github.com/coq/coq/blob/5eedc0b094c7aff4b10e89c547dcb79e1cc672c3/tools/CoqMakefile.in#L396-L401

JasonGross:
I have created #527 for the .v file part.  I am not sure how to install the .glob files, as it seems they are not even fully generated by the makefile (different directories overwrite each others' .glob files it seems).  Why does compcert invoke `-dump-glob` explicitly rather than allowing coq to dump the glob file alongside the .v files?  This code dates back to the initial git import of compcert.

xavierleroy:
> Note that coq_makefile-made makefiles install the .v and .glob files:

I dodn't know about this convention.  Flocq doesn't install .glob files, for instance.

>  I am not sure how to install the .glob files, as it seems they are not even fully generated by the makefile (different directories overwrite each others' .glob files it seems).

I'd be surprised.  Some target-dependent modules have multiple implementations in different directories, but only one is compiled.

> Why does compcert invoke -dump-glob explicitly rather than allowing coq to dump the glob file alongside the .v files?

A long time ago, .glob files were not generated by default, so an explicit `-dump-glob <filename>.glob` was needed.  Putting the .glob files in a separate directory would reduce clutter.  But nowadays this is all moot, esp. since we now have much clutter to begin with (.vok, .vos, .aux, ...).

See #529 for a proposed fix.

JasonGross:
> > Note that coq_makefile-made makefiles install the .v and .glob files:
> 
> I dodn't know about this convention. Flocq doesn't install .glob files, for instance.
> 

Flocq seems to have [a separate `install-glob` target](https://gitlab.inria.fr/flocq/flocq/-/blob/master/Remakefile.in?ref_type=heads#L145-149).  But indeed the opam package does not install globs, I will prepare a PR

JasonGross:
> Some target-dependent modules have multiple implementations in different directories, but only one is compiled.

Ah, I see, that makes sense.  I was assuming things based just on the directory structure.



[...]

PR #527, opened 2024-10-18
install .v files alongside .vo files

JasonGross:
This is partial work towards #526, and should enable the bug minimizer to at least sometimes minimize files from compcert by reconstructing the .glob files.  I am not sure how to install the .glob files, as it seems they are not even fully generated by the makefile (different directories overwrite each others' .glob files it seems).


xavierleroy (approved):
Agreed.  Thanks!

PR #528, opened 2024-10-18
Add -ignore-menhir-version

JasonGross:
This will allow the opam package to stop patching configure

### DIFF
--- a/configure
+++ b/configure
@@ -30,6 +30,7 @@ clightgen=false
 install_coqdev=false
 ignore_coq_version=false
 ignore_ocaml_version=false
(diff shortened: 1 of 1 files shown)


xavierleroy:
I'm sorry but I'm not comfortable with using old versions of Menhir with CompCert.  Just update the OPAM package dependencies to require 20200624 or later.

JasonGross:
What about a version that is `-allow-unreleased-menhir` only?  This is so that the dev version of coq-compcert no longer needs to patch compcert to work with the dev version of menhir.

PR #529, opened 2024-10-21
Build `.glob` files at the default place and install them

xavierleroy:
See #526 for motivations.

- Build `.glob` files in the same directory as the source `.v` file (Coq default)
- Install `.glob` files along with `.vo` and `.v` files

Fixes: #526


JasonGross (approved):
Thank you!

PR #530, opened 2024-10-21
Add -allow-unreleased-menhir

JasonGross:
This will allow https://github.com/coq/opam/tree/master/extra-dev/packages/coq-compcert/coq-compcert.dev to drop the patch that makes `./configure` compatible with unreleased / dev versions of menhir.

### DIFF
--- a/configure
+++ b/configure
@@ -30,6 +30,7 @@ clightgen=false
 install_coqdev=false
 ignore_coq_version=false
 ignore_ocaml_version=false
+allow_unreleased_menhir=false
 library_Flocq=local
 library_MenhirLib=local
 
@@ -101,6 +102,7 @@ Options:
   -install-coqdev      Also install the Coq development (implied by -clightgen)
   -ignore-coq-version  Accept to use experimental or unsupported versions of Coq
   -ignore-ocaml-version Accept to use experimental or unsupported versions of OCaml
+  -allow-unreleased-menhir Accept to use experimental / unreleased versions of Menhir
 '
 
 #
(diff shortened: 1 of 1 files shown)


xavierleroy:
I must have missed something.  Why do we need to support unreleased / dev versions of Menhir?  They are unreleased for a reason...

JasonGross:
Hm, I guess I was poking at things I don't understand, sorry. 

@MSoegtropIMC added such a patch in https://github.com/coq/opam/pull/1536, saying
> The patched to allow unreleased menhir versions should stay permanently in the dev package, since it is not desirable to merge it upstream (a well known menhir version is preferable there).

and

>  FYI : this is required for the Coq platform master branch

I thought, upon seeing only the code, without looking into its history, it would make sense to add a flag here.  (I originally thought it was required for the master version of Coq, but I guess not.)

I guess the reason this is required for Coq Platform is that the dev switch / master version requires that all packages it installs be compatible, and the dev version of coq-menhirlib is otherwise incompatible with CompCert.  @MSoegtropIMC is this correct?

xavierleroy:
No problem!  Thank you and @MSoegtropIMC for the clarification.

MSoegtropIMC:
> I guess the reason this is required for Coq Platform is that the dev switch / master version requires that all packages it installs be compatible, and the dev version of coq-menhirlib is otherwise incompatible with CompCert. @MSoegtropIMC is this correct?

Yes. if we build Coq Platform dev all packages are dev packages and CompCert needs to compile with this dev version of menhir. But it is not desirable to have this outside of the dev packages, so I think it could/should stay in the dev opam package forever.

PR #532, opened 2024-10-29
Use GNU make's wildcard expansion instead of the shell's, because the…

monniaux:
… shell will leave a export/*.glob item in there if there are no export/*.glob files.

The Makefile only works with gmake anyway.
It would be possible to use option nullglob in bash, but many systems now use dash.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ documentation: $(FILES)
 	mkdir -p doc/html
 	rm -f doc/html/*.html
 	coq2html -d doc/html/ -base compcert -short-names \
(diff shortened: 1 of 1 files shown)


monniaux:
And this script also works with MacOS Make.

xavierleroy:
In general I like the use of `make` functions instead of shell scripts, e.g. `$(foreach ...)` instead of `for`.  The problem with `$(wildcard ...)` in commands is that it can generate very long command lines that exceed Windows' limits.  We ran into this problem with the `rm` commands in `make clean`; that's why they still use shell globbing.

I'm going to push a different fix for issue  #533.

Issue #531, opened 2024-10-29
static qualifier in array parameter declaration


monniaux:
C99 §6.7.5.3-7 states that `static` in an array parameter declaration means
> If the keyword static also appears within the [ and ] of the
> array type derivation, then for each call to the function, the value of the corresponding
> actual argument shall provide access to the first element of an array with at least as many
> elements as specified by the size expression.

Thus the following should be accepted:
```
int foo5(int t[static 1]) {
  return t[0];
}

int foo6(int t[const static 1]) {
  return t[0];
}
```

I suspect the simplest is to ignore this keyword. (Technically, one could use it to convey semantic information that the area pointed to should be accessible for a certain length, thus allowing some optimizations such as speculatively loading values, but this would require quite a bit of work…)

xavierleroy:
Tentative fix at #539.

PR #539, opened 2024-11-29
Support C99 array declarator syntax involving `static` and `*`

xavierleroy:
E.g. `int x[static 10]` or `int x [*]`.  Currently, treated like `int x[10]` and `int x[]` respectively.

Ref: https://en.cppreference.com/w/c/language/array ; C99 6.7.5.2, C11 6.7.6.2.

Fixes: #531

### DIFF
--- a/cparser/Parser.vy
+++ b/cparser/Parser.vy
@@ -649,8 +649,23 @@ direct_declarator:
 | decl = direct_declarator LBRACK RBRACK
     { let 'Cabs.Name name typ attr loc := decl in
       Cabs.Name name (Cabs.ARRAY typ [] None) attr loc }
-(*| direct_declarator LBRACK ... STATIC ... RBRACK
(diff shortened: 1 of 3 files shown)


monniaux:
Fails line 7:
```
void foo(int [5]);
void fooc(int x[const 5]);
void foos(int x[static 5]);
void foov(int x[volatile 5]);
void foor(int x[restrict 5]);
void fooc(int [const 5]);
void foos(int [static 5]);
void foov(int [volatile 5]);
void foor(int [restrict 5]);
void fooc(int (* const x));
void foos(int *x);
void foov(int * volatile x);
void foor(int * restrict x);
void fooc(int x[volatile 5])
```

xavierleroy:
The parsers needed updating in two places, "direct declarators" and "direct abstract declarators", adding to the already terrible combinatorics of this part of the C grammar.  Should be all right now.

PR #544, opened 2025-01-14
Use the --print-version command to robustly get the Coq / Rocq version.

ppedrot:
This has been around since Coq 8.6, and it's much better than parsing the human-readable output of `--version`.


ppedrot:
Ping @xavierleroy just in case.

xavierleroy (approved):
Thanks for the code. Looks good to me.

PR #547, opened 2025-03-21
Support for Rocq/Coq 9.0

xavierleroy:
This PR adds basic support for the Rocq prover version 9.0.

We still use the `coqc`, `coqtop`, etc, command-line tools, relying on the Coq 9.0 compatibility layer.

The main change is the explicit prefixing of imports from the standard library with `From Coq`.  At some later point this will become `From Stdlib`, but right now this would break compatibility with Coq 8.

### DIFF
--- a/.github/workflows/latest.yml
+++ b/.github/workflows/latest.yml
@@ -16,10 +16,10 @@ jobs:
       target: ${{ matrix.target }}
       os: linux
       jobs: 4
-      opamroot: /home/coq/.opam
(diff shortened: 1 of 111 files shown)


xavierleroy:
CI action `latest` not yet updated to use 9.0, see https://github.com/coq-community/docker-rocq/issues/10 .

xavierleroy:
The docker situation was resolved (thanks @erikmd !), so the `latest` CI action now uses Coq/Rocq 9.0.

PR #548, opened 2025-05-04
Require Znumtheory before using it

andres-erbsen:
For https://github.com/rocq-prover/stdlib/pull/136


xavierleroy (approved):
Thanks for the patch.  It's better to be explicit in Require, in all cases.

PR #557, opened 2025-07-28
Support x86 Indirect Branch Tracking.

xavierleroy:
This PR adds support for IBT, a control-flow integrity mechanism in the x86 architecture.  It adds `endbr64` instructions at function entry points, so that they can be called via a pointer, and a `notrack` prefix for jumps through jump tables.

These instructions are generated if the `-fcf-protection=branch` option is given on the command-line.  On BSD systems, `-fcf-protection=branch` is the default and `-fcf-protection=none` can be given to turn IBT off.

If `-fcf-protection=branch` is given, appropriate ELF annotations are added to object files so that the Linux linker and loader can activate IBT at run-time if all object files have been compiled with `-fcf-protection=branch`.  (OpenBSD ignores this annotation and activates IBT by default, except if the program was linked with `-z nobtcfi`.)

Few Intel processors and no AMD processors support IBT in hardware.  On other processors, the special IBT instructions are treated like NOPs.  I have not been able to test that this PR actually implements IBT.  I just tested that the generated code works fine on non-IBT-enabled processors and carries the ELF annotations that Linux expects.


xavierleroy:
Also: IBT support is for 64-bit x86 only.  Intel processors support IBT in 32-bit mode as well, but I don't think it's worth supporting in CompCert.

PR #572, opened 2026-01-27
Remove support for x86 32 bits

xavierleroy:
The 32-bit x86 port of CompCert is no longer useful, now that all x86 processors made in the last 15 years are 64-bit processors and all x86 OS are 64-bit by default and increasingly stop supporting 32-bit applications.  There are a few legacy embedded systems built around 32-bit x86 processors, but they lack SSE2 support, which CompCert requires to implement FP arithmetic.

At the  same time, 32-bit x86 support required unnatural changes in CompCert, to accommodate peculiarities of the x86-32 ABI: FP function results returned on the FP stack, reduced alignment for `double` and `long long`, etc.  Removing this code will make CompCert easier to support in the future.


xavierleroy:
It turns out there remains some significant uses of the x86-32 back-end.  Plus, https://github.com/AbsInt/CompCert/commit/6a4d2584fb90d9ebfc43590890c45f1e09700f5d fixes a big infelicity of this port (concerning 4-alignment of 64-bit integers).  Closing.

PR #581, opened 2026-04-18
Enforce minimal safe alignment on global variables

xavierleroy:
Currently, global variables are aligned according to their C types, e.g. `char c[4];` has alignment 1.

However, the CompCert C semantics make it possible to cast `c` to `int *` and dereference this pointer, which requires 4-alignment of `c` to be on the safe side.

This PR makes sure that the alignment of a global variable is at least the minimal safe alignment for its size, as defined in `Memdata.min_safe_alignment`.  This ensures that any memory access (at offset 0) that fits within the bounds of the variable is aligned.

Without `_Alignas` modifiers, the alignment of a C type is always <= the minimal safe alignment of the size of the type.  However, `_Alignas` modifiers can increase the C alignment beyond the minimal safe alignment.  We use the greater of the two alignments.

### DIFF
--- a/backend/Inlining.v
+++ b/backend/Inlining.v
@@ -12,7 +12,7 @@
 
 (** RTL function inlining *)
 
-Require Import Coqlib Wfsimpl Maps Errors Integers.
+Require Import Coqlib Wfsimpl Maps Errors Integers Memory.
 Require Import AST Linking.
 Require Import Op Registers RTL.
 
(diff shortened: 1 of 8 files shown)


xavierleroy:
GCC and Clang also increase alignment of global variables beyond the natural alignment of their C types.  Consider:
```
char a[1];
char b[2];
char c[4];
char d[8];
char e[16];
```
I observed the following alignments:

| Compiler and target  | a | b | c | d | e |
| ------  | - | - | - | - | - |
| x86-64, gcc 13.3 | 1 | 1 | 1 | 8 | 16 |
| x86-64, clang 20.1 | 1 | 1 | 1 | 1 | 1 |
| aarch64, clang, macos | 1 | 2 | 4 | 8 | 16 |
| aarch64, gcc, linux | 8 | 8 | 8 | 8 | 8 |
| CompCert, before | 1 | 1 | 1 | 1 | 1 |
| CompCert, after this PR | 1 | 2 | 4 | 8 | 8 |

PR #584, opened 2026-06-14
x86 / Win64 ABI: fix inconsistency on callee-save XMM registers

xavierleroy:
As reported by Christos Papakonstantinou (Cantina Security),  `is_callee_save` was treating the XMM6 and XMM7 registers as caller-save, while they are callee-save in the Win64 x86 ABI.


m-schmidt:
Looks good to me.

PR #587, opened 2026-07-10
Full support for Rocq 9.2

xavierleroy:
This PR addresses the new warnings introduced by Rocq 9.2, either by rewriting the Rocq code in the new recommended way, or by silencing the warning.

It also changes the configure and build system so that the `rocq` command is used directly when available.  This makes it possible to build CompCert if Rocq is installed but the Coq-compatibility package is not.

It also declares OCaml 5 as officially supported, since we had no problems in testing and OCaml 5 is now fully supported by Rocq.

The PR is best reviewed commit per commit.


common/Determinism.v
```diff
@@ -504,8 +504,8 @@ Section WORLD_SEM.
 Variable L: semantics.
 Variable initial_world: world.
 
-Notation "s #1" := (fst s) (at level 9, format "s '#1'") : pair_scope.
-Notation "s #2" := (snd s) (at level 9, format "s '#2'") : pair_scope.
+Notation "s #1" := (fst s) (at level 1, format "s '#1'") : pair_scope.
+Notation "s #2" := (snd s) (at level 1, format "s '#2'") : pair_scope.
 Local Open Scope pair_scope.
```
bschommer, line 507: These can be removed completely since they are already available from Globalenvs.v

xavierleroy: Well spotted!  I don't know yet if I'd rather have those notations local, or make them globally available from Coqlib.v.  Will think about it.

xavierleroy: Finally, I moved the notations to Coqlib, so that they can be used in other files in the future.

Makefile
```diff
@@ -51,33 +51,53 @@ endif
...
+  -w -deprecated-from-Coq \
+  -w -register-all \
+  -w -notation-for-abbreviation \
+  -w -implicit-create-rewrite-hint-db \
+  -w -deprecated-end-tac
 
-cparser/Parser.vo: COQCOPTS += -w -deprecated-instance-without-locality
-MenhirLib/Interpreter.vo: COQCOPTS += -w -undeclared-scope
+# deprecated-instance-without-locality:
+#    warning introduced in 8.14
```
SkySkimmer, line 85: and removed in 8.18, with the default locality changing

xavierleroy: Duly noted, thanks!

PR #588, opened 2026-07-11
String literals in asm files: use asm quoting conventions, not OCaml's

xavierleroy:
Currently, we use the `%S` format of `Printf.printf` to emit some string literals in assembly files: filenames in `.file` directives, and `.asciz` directives for DWARF debug information.

(String literals coming from C source files follow a different path that doesn't use `%S`.  They are not affected by the problem fixed in this commit.)

The `%S` format produces an OCaml string literal, which can be misinterpreted by the GNU and Diab assemblers.  In particular, OCaml uses decimal escapes `\ddd` (three decimal digits) while assemblers, like C, use octal escapes `\ooo` (three octal digits).  Consequently, a "vertical tab" character (ASCII 0x0B) is printed as `\011`, which is interpreted as "horizontal tab" (ASCII 0x09) by the assembler.

This commits introduces `PrintAsmaux.quote_string` to produce a string literal in asm / C syntax, with octal escape sequences, and uses it in assembler printers in place of `%S`.

### DIFF
--- a/backend/PrintAsmaux.ml
+++ b/backend/PrintAsmaux.ml
@@ -300,6 +300,24 @@ let string_exists predicate s =
     exists (i + 1)
   in exists 0
 
+(** Quote the given string, using C string literal syntax.
+    The resulting string is enclosed in double quotes.
+    Double quotes and backslash are backslash-quoted.
+    Control characters (incl. newline and tabs) are octal-encoded. *)
+
+let quote_string s =
+  let b = Buffer.create (String.length s + 10) in
+  Buffer.add_char b '"';
+  String.iter
+    (fun c ->
+      match c with
+      | '"' | '\\' -> Buffer.add_char b '\\'; Buffer.add_char b c
+      | '\x00' .. '\x1F' -> Printf.bprintf b "\\%03o" (Char.code c)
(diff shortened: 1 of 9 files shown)


backend/PrintAsmaux.ml
```diff
@@ -300,6 +300,24 @@ let string_exists predicate s =
...
+    Control characters (incl. newline and tabs) are octal-encoded. *)
+
+let quote_string s =
+  let b = Buffer.create (String.length s + 10) in
+  Buffer.add_char b '"';
+  String.iter
+    (fun c ->
+      match c with
+      | '"' | '\\' -> Buffer.add_char b '\\'; Buffer.add_char b c
+      | '\x00' .. '\x1F' -> Printf.bprintf b "\\%03o" (Char.code c)
```
m-schmidt, line 307: What about quoting at least newline and tab as "\n" and "\t" instead of octal for readability?

xavierleroy: Yes, let's do that.  Commit fcda834.

m-schmidt: Looks good. Thanks!

PR #589, opened 2026-07-15
Adapt to rocq-prover/rocq#22271

proux01:
Adapt to https://github.com/rocq-prover/rocq/pull/22271

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,10 +86,14 @@ MenhirLib/%.vo: COQCOPTS+=-w -deprecated-syntactic-definition -w -deprecated-sin
 #    warning introduced in 8.20, no alternative before 8.20
 # deprecated-from-Coq:
 #    see above
+# unknown-option:
+#    the "Set Extraction Prefix" command was introduced in 9.4
+
 COQEXTRACTOPTS ?= \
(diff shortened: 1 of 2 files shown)


SkySkimmer:
Should be ready to merge

xavierleroy:
Thanks for the early notice.  One more warning to silence in the Makefile...  Merging after CI completes.

Makefile
```diff
@@ -106,10 +86,14 @@ flocq/%.vo MenhirLib/%.vo: COQCOPTS+=\
 #    warning introduced in 8.20, no alternative before 8.20
 # deprecated-from-Coq:
 #    see above
+# unknown-option:
+#    the "Set Extraction Prefix" command was introduced in 9.4
```
SkySkimmer, line 90: wouldn't it make more sense to silence the warning locally?

xavierleroy: Well, this applies only to the extraction/extraction.v file, not the other Rocq source files, so it's rather local already.  Plus, I like to put all the warning management in a single file (the Makefile), in the hope that it makes it easier to remove warnings when we drop support for old versions.

PR #591, opened 2026-08-05
Makefile: use an explicit path for extraction/extraction.v

JasonGross:
`make extraction` passes `extraction/extraction.v` to `-load-vernac-source`. That path is implicit, so Rocq resolves it against the load path instead of treating it as a filename, and any installed package with an `extraction` subdirectory shadows CompCert's own file. CertiCoq installs `user-contrib/CertiRocq/extraction/extraction.v`, so when it is present the build fails with `Cannot find a physical path bound to logical path Glue.glue`. Prefixing `./` makes the path explicit.


xavierleroy (approved):
Well spotted, thanks !

PR #594, opened 2026-08-19
Register allocation: try aggressive coalescing first

xavierleroy:
The IRC graph coloring algorithm is often too prudent with the coalescing of moves, leaving moves unchanged by fear that coalescing them could cause more spilling.

As an experiment, this PR runs a first round of coloring with aggressive coalescing of moves. If this causes spilling, it retries coloring with iterated coalescing, like before.

This should result in fewer moves for functions that can run entirely in registers.  For other functions, there should be no difference with the current coloring strategy.  The only possible downside is increased compilation time because of the extra round of coloring.

I haven't tried to evaluate the impact on code size and execution times yet.

### DIFF
--- a/backend/IRC.ml
+++ b/backend/IRC.ml
@@ -669,7 +669,7 @@ let combine g u v =
 
 (* Attempt coalescing *)
 
-let coalesce g =
+let coalesce forced_coalescing g =
   let m = DLinkMove.pick g.worklistMoves in
   let x = getAlias m.src and y = getAlias m.dst in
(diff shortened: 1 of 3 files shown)


xavierleroy:
The performance impact seems slightly positive, but there's a one case where aggressive coalescing increases code size and possibly execution time: if it forces the use of callee-save registers that would not be used if IRC-style coalescing was performed.  The heuristic for the first round could be changed  to "no spilling and no callee-save register used", but then it would apply much less often.

PR #595, opened 2026-08-23
x86-64 `Pjmptbl`: normalize the argument before indexing into the jump table

xavierleroy:
The index argument of `Pjmptbl` is an unsigned 32-bit integer, not a 64-bit  integer, so it must be converted to 64 bits (by zeroing the top 32 bits) before being used as an index into the jump table.

Issue reported by Christos Papakonstantinou (Cantina Security).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ Bug fixes:
 - Thread the current typing environment through `Elab.elab_initializer`.
 - Printing of assembly files: quote command-line arguments when needed (#586)
 - Printing of assembly files: revised string quoting in debugging information (#588)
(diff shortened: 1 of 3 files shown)


xavierleroy:
It turns out that RISC-V 64 bits has a similar issue, although much less serious: the 32-bit index argument is implicitly extended to a signed 64-bit integer.  So, in the unlikely case where the jump table has more than 2^31 entries, we're addressing the wrong entry.  Commit bc5a65f proposes to fix this by zero-extending the index while multiplying it by 4.  It adds one shift-right instruction to the code generated for `Pbtbl`.  The alternative would be to reject jump tables with more than 2^31 entries.

xavierleroy:
The RISC-V fix was overkill: we're generating worse code for a case that never happens in practice.  I just added a run-time assertion that the jump table has at most 2^31 entries, which guarantees that the index has the correct 64-bit value.

PR #596, opened 2026-09-14
`__builtin_memcpy_aligned` for large blocks

xavierleroy:
The expansion of  `__builtin_memcpy_aligned` currently masks some of the high bits of the size argument off, resulting in partial copies for very large blocks (size >= 2^30, 2^34 or 2^36, depending on the target architecture). 

This is very unlikely to occur in practice, but this PR fixes this problem, making sure all bits of the size argument are honored.  

In passing, it also changes the expansion of `__builtin_memcpy_aligned` on x86-64 to take advantage of hardware optimizations on recent x86 processors, and reduces the alignment used to at most 8, which is the largest alignment supported by the CompCert back-end.

Issue (on x86-64) reported by Christos Papakonstantinou (Cantina Security).


xavierleroy:
Note that `__builtin_memcpy_aligned` with large sizes slow down value analysis and neededness analysis tremendously.  So, this can only be tested with `-O0`.

PR #598, opened 2026-09-21
Recognize dense `switch` statements involving negative and positive cases.

xavierleroy:
Currenty, `switch` statements with negative and positive cases, such as
```
switch (n) { case -1: ...  case 0: ... case 1: case 2: case 3: case 4: case 5: ... }
```
are never recognized as dense and never compiled to a jump table.  That's because case labels are treated as unsigned, hence any switch involving negative  and positive numbers has a very large span and looks very sparse.

This PR changes the computation of the span of a switch to treat labels as signed numbers.   Owing to the joys of modulo arithmetic, very little else needs to be changed: the validation of jump tables is relaxed to allow wrap around at 0, which is semantically safe; the translation of decision trees to CminorSel code is unchanged.


xavierleroy:
I also considered treating `case` labels as signed integers in all languages from CompCert C to Cminor, but that's a bigger change and there is no pressing need.

